Issue #33, opened 2020-01-08
Add pre-commit hook for client linting


AlmightyYakob:
This is ideally something similar to the current pre-commit hooks in place for python code, but would instead lint the `client` directory (equivalent to running `yarn lint`). This also may not be feasible, so determining that is also part of this issue.

waxlamp:
As part of this, please update the CONTRIBUTING.md file and developer documentation to explain how this works.

PR #46, opened 2020-03-16
Add CONTRIBUTING.md file

### LINKED ISSUE #28: Write a CONTRIBUTING.md file
Mention use of typescript. Include these rules:
- don't use `any` just to satisfy the compiler
- put semantically rich types into `types.ts`, and put auxiliary or convenience types in the files that use them

waxlamp:
Closes #28 

If we're happy with this file, I'll open parallel PRs (with the same CONTRIBUTING.md file) in multinet-server and multinetjs.


waxlamp:
> Should we add a section about code style?

I thought about this but wasn't sure what to even say. Do we have a name for the code style we use? We do have linters but I wasn't sure if this was the right place to get into that level of tooling (even though maybe other projects do discuss that stuff here).

I know @AlmightyYakob wants to set up commit hooks to prevent bad code from being committed; maybe once that's fully in place, we could update this with more detailed guidance?

> How about an issue template?

At this stage I'm still ok with free-form issue reports, in the interest of lowering the barrier to entry. Issue templates would be valuable if we were, say, swamped with vague issue reports that were not easily reproducible due to browser differences, etc. Since we're not there yet, I'm ok with keeping things easier for contributors.

JackWilb:
> > Should we add a section about code style?
> 
> I thought about this but wasn't sure what to even say. Do we have a name for the code style we use? We do have linters but I wasn't sure if this was the right place to get into that level of tooling (even though maybe other projects do discuss that stuff here).
> 
> I know @AlmightyYakob wants to set up commit hooks to prevent bad code from being committed; maybe once that's fully in place, we could update this with more detailed guidance?

Works for me. Should we open an issue

> > How about an issue template?
> 
> At this stage I'm still ok with free-form issue reports, in the interest of lowering the barrier to entry. Issue templates would be valuable if we were, say, swamped with vague issue reports that were not easily reproducible due to browser differences, etc. Since we're not there yet, I'm ok with keeping things easier for contributors.

That's fine by me, too. The only thing I'd add is that it might be easier to do that up front instead of later, but it doesn't seem like a big deal at the moment.

waxlamp:
> > > Should we add a section about code style?
> > 
> > 
> > I thought about this but wasn't sure what to even say. Do we have a name for the code style we use? We do have linters but I wasn't sure if this was the right place to get into that level of tooling (even though maybe other projects do discuss that stuff here).
> > I know @AlmightyYakob wants to set up commit hooks to prevent bad code from being committed; maybe once that's fully in place, we could update this with more detailed guidance?
> 
> Works for me. Should we open an issue?
> 

#33 

> > > How about an issue template?
> > 
> > 
> > At this stage I'm still ok with free-form issue reports, in the interest of lowering the barrier to entry. Issue templates would be valuable if we were, say, swamped with vague issue reports that were not easily reproducible due to browser differences, etc. Since we're not there yet, I'm ok with keeping things easier for contributors.
> 
> That's fine by me, too. The only thing I'd add is that it might be easier to do that up front instead of later, but it doesn't seem like a big deal at the moment.

Why would it be easier to do up front?

JackWilb:
> > > > How about an issue template?
> > > 
> > > 
> > > At this stage I'm still ok with free-form issue reports, in the interest of lowering the barrier to entry. Issue templates would be valuable if we were, say, swamped with vague issue reports that were not easily reproducible due to browser differences, etc. Since we're not there yet, I'm ok with keeping things easier for contributors.
> > 
> > 
> > That's fine by me, too. The only thing I'd add is that it might be easier to do that up front instead of later, but it doesn't seem like a big deal at the moment.
> 
> Why would it be easier to do up front?

If we have consistent contributors? Changing them over might be a pain point. I'm sure it's fine for now

waxlamp:
> Regarding the code style, I think the best route would be opening an issue and linking this PR, so that we remember to update this.

I left a link to this PR in #33; is that sufficient or do you think we need another issue to address some of this?

AlmightyYakob:
> I left a link to this PR in #33; is that sufficient or do you think we need another issue to address some of this?

I think that's good :+1:

PR #49, opened 2020-03-19
Fix extra space in random phrase

AlmightyYakob:


AlmightyYakob:
> Good catch. Did you happen to `git grep " '"` to look for others like this?

Yep, this was the only one I found.

PR #59, opened 2020-03-26
Add renaming workspace functionality

### LINKED ISSUE #18: Add ability to rename workspaces

AlmightyYakob:
Fixes #18.

There are some intentional issues with this PR, and will be fixed soon when we add Vuex, but since that's not in the project yet I've done workarounds.

@jtomeck The UI also needs to be improved, specifically the text-field in the app bar looks odd, and the save button is very far to the right. I played around with it a bit but couldn't get it to behave how I wanted, so I thought you could take a crack at it.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,7 +1,9 @@
 <template>
   <v-app id="app">
-    <sidebar />
-    <router-view :apps='apps' />
+    <!-- TODO: REMOVE THIS REF WHEN VUEX IS ADDED -->
+    <sidebar ref="sidebar" />
+    <!-- TODO: REMOVE THIS UPDATE WHEN VUEX IS ADDED -->
+    <router-view :apps='apps' @update="update"/>
   </v-app>
 </template>
 
@@ -21,6 +23,12 @@ export default Vue.extend({
   },
   created() {
     this.apps = require('./appregistry.json');
+  },
+  methods: {
+    // TODO: REMOVE THIS REF AND METHOD WHEN VUEX IS ADDED
+    update() {
+      this.$refs.sidebar.refresh();
+    },
   },
 });
 </script>
--- a/src/App.vue
+++ b/src/App.vue
@@ -27,7 +27,8 @@ export default Vue.extend({
   methods: {
     // TODO: REMOVE THIS REF AND METHOD WHEN VUEX IS ADDED
     update() {
-      this.$refs.sidebar.refresh();
+      const sidebar = this.$refs.sidebar as any;
+      sidebar.refresh();
     },
   },
 });
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "axios": "^0.18.1",
     "core-js": "^2.6.5",
     "material-design-icons-iconfont": "^5.0.1",
-    "multinet": "0.10.0",
+    "multinet": "0.12.0",
     "vue": "^2.6.10",
     "vue-gtag": "^1.2.1",
     "vue-router": "^3.0.2",
--- a/src/components/DeleteGraphDialog.vue
+++ b/src/components/DeleteGraphDialog.vue
@@ -10,13 +10,13 @@
         <template v-slot:activator="{ on: tooltip }">
           <v-scroll-x-transition>
             <v-btn
+              class="ml-1"
               icon
               small
-              text
               @click="dialog.click"
               v-on="tooltip"
               >
-              <v-icon color="red accent-3" size="22px">delete_sweep</v-icon>
+              <v-icon color="red accent-3" size="20px">delete_sweep</v-icon>
             </v-btn>
           </v-scroll-x-transition>
         </template>
--- a/src/components/DeleteTableDialog.vue
+++ b/src/components/DeleteTableDialog.vue
@@ -10,13 +10,13 @@
         <template v-slot:activator="{ on: tooltip }">
           <v-scroll-x-transition>
             <v-btn
+              class="ml-1"
               icon
               small
-              text
               @click="dialog.click"
               v-on="tooltip"
               >
-              <v-icon color="red accent-3" size="22px">delete_sweep</v-icon>
+              <v-icon color="red accent-3" size="20px">delete_sweep</v-icon>
             </v-btn>
           </v-scroll-x-transition>
         </template>
--- a/src/components/DownloadDialog.vue
+++ b/src/components/DownloadDialog.vue
@@ -12,11 +12,10 @@
             <v-btn
               icon
               small
-              text
               @click="dialog.click"
               v-on="tooltip"
             >
-              <v-icon color="primary" size="22px">save_alt</v-icon>
+              <v-icon color="primary" size="20px">save_alt</v-icon>
             </v-btn>
           </v-scroll-x-transition>
         </template>
--- a/src/components/FileUploadForm.vue
+++ b/src/components/FileUploadForm.vue
@@ -167,6 +167,8 @@ export default Vue.extend({
       } catch (err) {
         if (err.status === 409) {
           this.tableCreationError = `Table "${fileName}" already exists`;
+        } else if (err.status === 415) {
+          this.tableCreationError = 'Data could not be read as CSV';
         } else {
           this.tableCreationError = 'Unknown error; please see developer console for details';
           throw err;
--- a/src/components/GraphDialog.vue
+++ b/src/components/GraphDialog.vue
@@ -6,14 +6,12 @@
     <template v-slot:activator="{ on }">
       <v-btn
         id="add-graph"
-        class="new-button"
         color="blue darken-2"
-        fab
-        dark
+        icon
         medium
         v-on="on"
       >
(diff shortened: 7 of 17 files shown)


jtomeck:
Sounds good @AlmightyYakob! I'll Probably have time to take a look on Monday if that's alright? I have a quick deliverable due end of day today. Just wanted to let you know so you didn't think I missed the review request.

AlmightyYakob:
> I'll Probably have time to take a look on Monday if that's alright?

Yeah sounds good, thanks for letting me know!

src/App.vue
```diff
@@ -22,6 +24,12 @@ export default Vue.extend({
   created() {
     this.apps = require('./appregistry.json');
   },
+  methods: {
+    // TODO: REMOVE THIS REF AND METHOD WHEN VUEX IS ADDED
+    update() {
+      this.$refs.sidebar.refresh();
```
waxlamp: You'll need to get around the typechecker here, by doing something like
```suggestion
      (this.$refs.sidebar as any).refresh();
```

AlmightyYakob: See 9ab3a2e

AlmightyYakob:
@waxlamp I've addressed the following:

> * hitting enter should be the same as clicking on the save button
> * hitting escape should be the same thing as clicking the cancel button
> * if the user deletes the whole text field, saving should have no effect (or the save icon should become disabled, leaving canceling as the only option)
> * the disk icon is so hokey--maybe @jtomeck can come up with an alternative (as simple as a checkmark to contrast against the X that is used for "cancel")

Another thought I had: While updating the name in place would be cool if we could get it to behave properly, it'd also be at lot easier to just have all of this functionality in a dialog. Let me know what you guys think.

---

Regarding this conern:

> * we may need some validation on workspace names. We already check for whether a workspace by the same name exists, but checking for a blank name or a name consisting of only whitespace or something might be reasonable

I haven't made that change yet, but I think it would be within the scope of this PR to do so. However I'm okay doing it as a separate issue as well. Let me know what you think.

waxlamp:
> @waxlamp I've addressed the following:
> 
> > * hitting enter should be the same as clicking on the save button
> > * hitting escape should be the same thing as clicking the cancel button
> > * if the user deletes the whole text field, saving should have no effect (or the save icon should become disabled, leaving canceling as the only option)
> > * the disk icon is so hokey--maybe @jtomeck can come up with an alternative (as simple as a checkmark to contrast against the X that is used for "cancel")
> 
> Another thought I had: While updating the name in place would be cool if we could get it to behave properly, it'd also be at lot easier to just have all of this functionality in a dialog. Let me know what you guys think.
> 

I think we're pretty much there--the checkmark looks great, the enter/escape functionality works like a dream, and really all that's left (besides validation; see below) is for @jtomeck to work his magic on the styles.

> Regarding this conern:
> 
> > * we may need some validation on workspace names. We already check for whether a workspace by the same name exists, but checking for a blank name or a name consisting of only whitespace or something might be reasonable
> 
> I haven't made that change yet, but I think it would be within the scope of this PR to do so. However I'm okay doing it as a separate issue as well. Let me know what you think.

It's definitely in scope.

I think a good approach here is to fail validation for an all-whitespace name with `Workspace name cannot be blank`.

I don't know quite what to do for workspace names that begin with whitespace--I kind of think that should just not be allowed, as I feel like it could create chaos down the line. I feel similarly about trailing whitespace in workspace names, for the sake of simplicity at the very least. What do you think?

AlmightyYakob:
> I think a good approach here is to fail validation for an all-whitespace name with `Workspace name cannot be blank`.

Agree

> I don't know quite what to do for workspace names that begin with whitespace--I kind of think that should just not be allowed, as I feel like it could create chaos down the line. I feel similarly about trailing whitespace in workspace names, for the sake of simplicity at the very least. What do you think?

I also think it shouldn't be allowed. I can make these changes.

AlmightyYakob:
@waxlamp I've added the features you requested, namely not allowing white space at the beginning or end of the name. Also, if you try to rename a workspace to the name of an existing workspace, an error appears to the user. Let me know what you think of these changes.

jtomeck:
> the save button is very far to the right

I'm not sure we need a save button, but if you and @waxlamp disagree, I can try and figure something out.

> it'd also be at lot easier to just have all of this functionality in a dialog.

I kind of like editing in place better, because it's much more slick and feels more part of the UI. I don't really like using dialogs for things that only have 1 field. If I had the skills when I created the create workspace dialog/button I probably would have done things differently, where clicking the button would add an item in the workspaces list with an active text field so you can type in the name and hit enter to commit it (and auto switch to the new workspace page).

That said, since I am thinking the rename feature might be good as an action in the sidebar when you hover over a workspace as well, so I guess there's a few options:

1. Keep as is and have the header bar be the only place to rename a workspace
2. Add an edit action on hover to the sidebar (no dialog) to rename workspaces. (With this option we could keep or even get rid of the edit in the header bar, since the workspaces sidebar is always visible)
3. Add an edit action on hover to the sidebar to pop out a dialog for renaming (again we could have it in both places or just the sidebar).

@waxlamp @AlmightyYakob - thoughts on the novel I just wrote?

AlmightyYakob:
> 1. Keep as is and have the header bar be the only place to rename a workspace
> 2. Add an edit action on hover to the sidebar (no dialog) to rename workspaces. (With this option we could keep or even get rid of the edit in the header bar, since the workspaces sidebar is always visible)
> 3. Add an edit action on hover to the sidebar to pop out a dialog for renaming (again we could have it in both places or just the sidebar).
> 

I think for this PR, keeping the top bar as the only place to rename is the way to go, as we've already iterated on this feature a good amount, and the feature is fully functional after we fix up/remove the confirmation button. 

If we wanted the feature in both places, my inclination would be towards having a dialog, although not a whole modal, just like a small popover where the editing is happening. But I don't feel too strongly about this, I'll let @waxlamp weigh in.

AlmightyYakob:
Also regarding this:
> I'm not sure we need a save button, but if you and @waxlamp disagree, I can try and figure something out.

I don't feel strongly one way or the other. I'd say we could get away without one, and we might want to for the time being, but if we ever want to add more things to that top bar we're going to have to figure out the spacing up there anyway.

jtomeck:
> if we ever want to add more things to that top bar we're going to have to figure out the spacing up there anyway.

I think if we add things on the right, it will automatically resize properly since I believe it's done using flexbox. The reason the name field is so long is to account for super long names of workspaces/tables/graphs.

waxlamp:
> > 1. Keep as is and have the header bar be the only place to rename a workspace
> > 2. Add an edit action on hover to the sidebar (no dialog) to rename workspaces. (With this option we could keep or even get rid of the edit in the header bar, since the workspaces sidebar is always visible)
> > 3. Add an edit action on hover to the sidebar to pop out a dialog for renaming (again we could have it in both places or just the sidebar).
> 
> I think for this PR, keeping the top bar as the only place to rename is the way to go, as we've already iterated on this feature a good amount, and the feature is fully functional after we fix up/remove the confirmation button.
> 

I agree with keeping the edit action just in the top bar right now. I like the proposals for editing workspace names in the sidebar but I'd ask @jtomeck to file an issue for that just so we don't forget that we liked the idea.

The confirmation button could either be taken out entirely, or it could appear in a vertical stack with the cancel button to the left. Let's just do whatever is easier, unless there's a compelling reason for something more complex. What do you think is the best course of action, @jtomeck?

jtomeck:
I think let's ditch the confirm button and I'll see what I can do with the shifting of the text field when editing the name. I've got a 3pm meeting, but I'll jump in and take care of it afterwards.

AlmightyYakob:
@jtomeck It looks like the spacing fix you added is pushing the error messages too far down (or something). Is this something you would be able to address?

jtomeck:
> @jtomeck It looks like the spacing fix you added is pushing the error messages too far down (or something). Is this something you would be able to address?

Ah, yea sorry. I always forget to check the form validation messages when styling them. Sorry!

AlmightyYakob:
@jtomeck LGTM, now just need @waxlamp's approval

src/App.vue
```diff
@@ -1,7 +1,9 @@
 <template>
   <v-app id="app">
-    <sidebar />
-    <router-view :apps='apps' />
+    <!-- TODO: REMOVE THIS REF WHEN VUEX IS ADDED -->
+    <sidebar ref="sidebar" />
+    <!-- TODO: REMOVE THIS UPDATE WHEN VUEX IS ADDED -->
```
waxlamp: Could you open an issue for these and link the lines of code in the issue description?

src/App.vue
```diff
@@ -22,6 +24,13 @@ export default Vue.extend({
   created() {
     this.apps = require('./appregistry.json');
   },
+  methods: {
+    // TODO: REMOVE THIS REF AND METHOD WHEN VUEX IS ADDED
```
waxlamp: Same as above.

PR #63, opened 2020-03-27
Fix props declarations

AlmightyYakob:
Update the props definitions in various files to be specific.

### DIFF
--- a/src/views/WorkspaceDetail.vue
+++ b/src/views/WorkspaceDetail.vue
@@ -157,7 +157,7 @@
 </template>
 
 <script lang="ts">
-import Vue from 'vue';
+import Vue, { PropType } from 'vue';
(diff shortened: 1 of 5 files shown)


src/views/WorkspaceDetail.vue
```diff
@@ -177,7 +177,9 @@ export default Vue.extend({
     DeleteTableDialog,
     DownloadDialog,
   },
-  props: ['workspace', 'title'],
+  props: {
+    workspace: String as PropType<string>,
+  },
```
JackWilb: Did you lose `title` here? Or is it not needed any more?

AlmightyYakob: It's never used in the code so I removed it

Issue #37, opened 2020-02-14
Add VueX to manage application-wide state


waxlamp:

waxlamp:
@AlmightyYakob: @subdavis gained some insight into the use of VueX as the Vue conference he attended last week. We should pull him into a discussion about our use case and see what recommendations he might have for us.

subdavis:
Since ya'll are already using TS, I would kind of like to see how a `Vuex-lite` would work here.

I found this article, which gives a really simple overview of the concept: https://dev.to/cooperaustinj/using-vue-observable-as-a-state-store-37fa

I probably wouldn't follow that style though.  You could separate your action functions out into its own file and have the full power of typescript.

waxlamp:
> Since ya'll are already using TS, I would kind of like to see how a `Vuex-lite` would work here.
> 
> I found this article, which gives a really simple overview of the concept: https://dev.to/cooperaustinj/using-vue-observable-as-a-state-store-37fa

This looks like a really interesting approach. Are the main reasons to do it this way to (1) avoid the full complexity of VueX and (2) get full TS support?

> I probably wouldn't follow that style though. You could separate your action functions out into its own file and have the full power of typescript.

Does doing it the blog post way prevent you from getting full TS support? Would you instead define functions in a file, then re-export those into an object from a top-level file, so you sort of implicitly have a version of VueX namespaces?

And on a broader note: do you think that following this approach could have its own pitfalls in terms of not following the standard approach of just using VueX? In my opinion, not having full TS support is a big loss, but that needs to be balanced against the risk of doing things in an uncharted way. This "Vuex-lite" approach seems simple enough that the risk seems small to me, but I'm curious what you think.

Thanks @subdavis!

subdavis:
> Does doing it the blog post way prevent you from getting full TS support?

Oh no you're right.  That style would be fine.  I suppose the only significant difference between that style and individual functions is that each component could import exactly the functions it cares about.  I sort of like that anyway, because removing or modifying function names will be build-time errors whether you typescript or not.

> Are the main reasons to do it this way to (1) avoid the full complexity of VueX and (2) get full TS support?

That's what I'm expecting.

> And on a broader note: do you think that following this approach could have its own pitfalls in terms of not following the standard approach of just using VueX? In my opinion, not having full TS support is a big loss, but that needs to be balanced against the risk of doing things in an uncharted way. This "Vuex-lite" approach seems simple enough that the risk seems small to me, but I'm curious what you think.

[...]

PR #66, opened 2020-03-31
Wire up individual item deletion/download icons

### LINKED ISSUE #60: Hook up individual item delete and download buttons
This refers to providing the correct behavior implementation for the new UI elements added in #56.

waxlamp:
This PR provides action semantics for the item-level download and delete icons, for workspaces, networks, and tables.

To accomplish this more easily, I decided to respecialize the `ItemPanel` component into `TablePanel` and `NetworkPanel` components. Unfortunately, there are some (reasonable) limitations in Vue around accessing slot content via refs, and while using refs to manipulate child components like `v-dialog` instances is not a best practice, it seemed to be warranted in this case.

In the future, we may want to do the following:
- use actual best practices to handle the lifecycle of the various dialogs in use in these panels
- refactor the repeated logic between the two panels into a mixin

But I don't think of these as at all urgent. If we need to make changes to these panels, or add a third panel, then these may become more important.

For now, please mess around with these and see if I missed any corner cases in behavior. If these are good, I want to get this merged and conclude our usability sprint :tada:

Closes #60


JackWilb:
Looks to work pretty well, I'll dig into the code a little later. I noticed one thing so far:

When hitting delete on any table that is used in a network, the dialog pops up and then immediately changes to the "you can't delete that yet" dialog. Is there a way we can have it pop up with the right message right away?

waxlamp:
> When hitting delete on any table that is used in a network, the dialog pops up and then immediately changes to the "you can't delete that yet" dialog. Is there a way we can have it pop up with the right message right away?

Good catch. I am noticing this behavior on multinet.app as well, so I've opened #67 to address it, but it shouldn't block the review/merge of this PR.

waxlamp:
> Works well locally, requested a small change. Somes notes:
> 
> * When [vuex](https://github.com/multinet-app/multinet-client/issues/37) is added, it will clean up this portion of code a bit, because the list of networks/tables will be managed by the store.
> 

For sure.

> * Looking at what's necessary for this, I think `DeleteWorkspaceDialog` should contain just the dialog, and not the activator button or anything. This would make more sense as far as the structure, and would allow us to directly control the state of the dialog, and remove the need for the `singleSelected` variable.
> 

Let's definitely revisit this after the big VueX change lands. I might be missing something because even if we extract the activator icon from inside the dialog component (a good change in general, IMO), we would still need the singleSelected mechanism, or something like it, in order to properly keep track of both the checkbox state, and handle one-off delete/download actions.

> 
> I think all of this is for separate PRs though, so I'm okay with approving this once my comments are addressed.

:confetti_ball:

src/components/TablePanel.vue
```diff
@@ -59,93 +68,137 @@
...
     },
+
+    cleanup(deleted?: string[]) {
+      const {
+        singleSelected,
+      } = this;
+
+      if (singleSelected !== null) {
+        this.singleSelected = null;
+      }
```
JackWilb, line 194: Can you simplify this to just: `this.singleSelected = null`?

src/components/NetworkPanel.vue
```diff
@@ -0,0 +1,234 @@
...
+    },
+
+    cleanup(deleted?: string[]) {
+      const {
+        singleSelected,
+      } = this;
+
+      if (singleSelected !== null) {
+        this.singleSelected = null;
+      }
```
JackWilb, line 205: Same as the other.

src/components/TablePanel.vue
waxlamp: See 14dcbf1dbaf16da19aa4551782d5a6419fad6314.

src/components/NetworkPanel.vue
waxlamp: See 14dcbf1dbaf16da19aa4551782d5a6419fad6314.

src/components/TablePanel.vue
JackWilb: Not a huge deal, but I think you can also delete the 
```
     const {
        singleSelected,
      } = this;
```

waxlamp: Good catch. See f25502a.

PR #68, opened 2020-04-01
Add Vuex

### LINKED ISSUE #37: Add VueX to manage application-wide state

AlmightyYakob:
Draft PR that will close #37 once finished. Since this is such a big change, @waxlamp thought it would be good (and I agree) to have this as a draft PR, so that changes can be discussed and seen openly.

Should probably be merged after #69 if you want to test the functionality.

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -36,12 +36,14 @@ const {
       }
       return [];
     },
+
     edgeTables(state: State): string[] {
       if (state.currentWorkspace !== null && state.currentWorkspace.edgeTables) {
         return state.currentWorkspace.edgeTables;
       }
       return [];
     },
+
     graphs(state: State) {
       if (state.currentWorkspace !== null && state.currentWorkspace.graphs) {
         return state.currentWorkspace.graphs;
@@ -53,9 +55,11 @@ const {
     setWorkspaces(state, workspaces: string[]) {
       state.workspaces = workspaces;
     },
+
     setCurrentWorkspace(state, workspace: WorkspaceState) {
       state.currentWorkspace = workspace;
     },
+
     setEmptyWorkspace(state, name: string) {
       state.currentWorkspace = {
         name,
@@ -71,6 +75,7 @@ const {
       const workspaces = await api.workspaces();
       commit.setWorkspaces(workspaces);
     },
+
     async fetchWorkspace(context, workspace: string) {
       const { commit } = rootActionContext(context);
       commit.setEmptyWorkspace(workspace);
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -60,13 +60,8 @@ const {
       state.currentWorkspace = workspace;
     },
 
-    setEmptyWorkspace(state, name: string) {
-      state.currentWorkspace = {
-        name,
-        nodeTables: [],
-        edgeTables: [],
-        graphs: [],
-      };
+    unsetCurrentWorkspace(state) {
+      state.currentWorkspace = null;
     },
   },
   actions: {
@@ -78,7 +73,7 @@ const {
 
     async fetchWorkspace(context, workspace: string) {
       const { commit } = rootActionContext(context);
-      commit.setEmptyWorkspace(workspace);
+      commit.unsetCurrentWorkspace();
 
       const nodeTables = await api.tables(workspace, { type: 'node' });
       const edgeTables = await api.tables(workspace, { type: 'edge' });
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -13,6 +13,11 @@ interface WorkspaceSpec {
   graphs?: string[];
 }
 
+interface State {
+  workspaces: string[];
+  currentWorkspace: WorkspaceSpec | null;
+}
+
 const {
   store,
   rootActionContext,
@@ -21,8 +26,28 @@ const {
   moduleGetterContext,
 } = createDirectStore({
   state: {
-    workspaces: [] as string[],
-    currentWorkspace: null as WorkspaceSpec | null,
+    workspaces: [],
+    currentWorkspace: null,
+  } as State,
+  getters: {
+    nodeTables(state: State): string[] {
+      if (state.currentWorkspace !== null && state.currentWorkspace.nodeTables) {
+        return state.currentWorkspace.nodeTables;
+      }
+      return [];
+    },
+    edgeTables(state: State): string[] {
+      if (state.currentWorkspace !== null && state.currentWorkspace.edgeTables) {
+        return state.currentWorkspace.edgeTables;
+      }
+      return [];
+    },
+    graphs(state: State) {
+      if (state.currentWorkspace !== null && state.currentWorkspace.graphs) {
+        return state.currentWorkspace.graphs;
+      }
+      return [];
+    },
   },
   mutations: {
     setWorkspaces(state, workspaces: string[]) {
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -70,7 +70,6 @@ const {
   },
 });
 
-
 export default store;
 export {
   rootActionContext,
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -1,0 +1,10 @@
+import Vue from 'vue';
+import Vuex from 'vuex';
+
+Vue.use(Vuex);
+
+export default new Vuex.Store({
+  state: {
+    workspaces: [] as string[],
+  },
+});
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -1,11 +1,17 @@
 import Vue from 'vue';
-import Vuex from 'vuex';
-
+import Vuex, { Store } from 'vuex';
 import { createDirectStore } from 'direct-vuex';
 
 import api from '@/api';
 
 Vue.use(Vuex);
+
+interface WorkspaceSpec {
+  name: string;
+  nodeTables?: string[];
+  edgeTables?: string[];
+  graphs?: string[];
+}
 
 const {
   store,
@@ -16,16 +22,29 @@ const {
 } = createDirectStore({
   state: {
     workspaces: [] as string[],
+    currentWorkspace: null as WorkspaceSpec | null,
   },
   mutations: {
-    setWorkspaces(state, workspaces) {
+    setWorkspaces(state, workspaces: string[]) {
       state.workspaces = workspaces;
+    },
+    setCurrentWorkspace(state, workspace: WorkspaceSpec) {
+      state.currentWorkspace = workspace;
     },
   },
   actions: {
     async fetchWorkspaces({ commit }) {
       const workspaces = await api.workspaces();
       commit('setWorkspaces', workspaces);
+    },
+    async fetchWorkspace({ commit }, workspace: string) {
+      const nodeTables = await api.tables(workspace, { type: 'node' });
+      const edgeTables = await api.tables(workspace, { type: 'edge' });
+      const graphs = await api.graphs(workspace);
+
+      // console.log("nodeTables --->", nodeTables);
+
+      commit('setCurrentWorkspace', { name: workspace, nodeTables, edgeTables, graphs });
     },
   },
 });
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -86,6 +86,7 @@ export {
   moduleGetterContext,
 };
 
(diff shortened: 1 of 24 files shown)


AlmightyYakob:
@waxlamp @JackWilb This should be good for review. I don't know how you guys feel about closing that issue with this merge, as it's not used in all of the spots it could be, this just adds it's use for workspaces mainly.

src/store/index.ts
```diff
@@ -0,0 +1,91 @@
...
+  name: string;
+  nodeTables?: string[];
+  edgeTables?: string[];
+  graphs?: string[];
+}
+
+interface State {
+  workspaces: string[];
+  currentWorkspace: WorkspaceSpec | null;
+}
```
JackWilb, line 19: Maybe one or both of these would be best defined in a global types.ts?

src/store/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      commit.setWorkspaces(workspaces);
+    },
+    async fetchWorkspace(context, workspace: string) {
+      const { commit } = rootActionContext(context);
+      const nodeTables = await api.tables(workspace, { type: 'node' });
+      const edgeTables = await api.tables(workspace, { type: 'edge' });
+      const graphs = await api.graphs(workspace);
+
+      commit.setCurrentWorkspace({ name: workspace, nodeTables, edgeTables, graphs });
+    },
```
JackWilb, line 73: It might be worthwhile doing `commit.setCurrentWorkspace({ name: workspace, [], [], [] })`  first, because when you switch workspaces at the moment, there is a little lag where the old names show. Setting everything to empty will at least not show the old networks and tables as you're moving around

AlmightyYakob: Good catch, just tried it locally and it makes the UI feel much smoother. I'll push soon.

src/store/index.ts
AlmightyYakob: Good point, I think it makes sense to move `WorkspaceSpec` there, but I think `State` is fine where it is, as it's inherent to the store.

waxlamp:
> @waxlamp @JackWilb This should be good for review. I don't know how you guys feel about closing that issue with this merge, as it's not used in all of the spots it could be, this just adds it's use for workspaces mainly.

This PR does exactly what the issue asks, so it should definitely be closed with this being merged. If it helps to manage the incremental process of adopting VueX everywhere, then definitely open new issues, but that's not completely required--whether to use VueX in a given location for a given reason is probably more art than science so it doesn't necessarily fit in with systematically closing issue to get it done.

waxlamp:
> I do think we should probably leave the issue open until we've refactored the rest of the app to use vuex.

It's a tall order and we have actual feature work we need to do (some of which only depends on VueX being available) so let's aim to merge this sooner rather than later and send followup PRs as needed to incrementally improve VueX usage.

src/store/index.ts
```diff
@@ -0,0 +1,87 @@
...
+      state.workspaces = workspaces;
+    },
+    setCurrentWorkspace(state, workspace: WorkspaceSpec) {
+      state.currentWorkspace = workspace;
+    },
+  },
+  actions: {
+    async fetchWorkspaces(context) {
+      const { commit } = rootActionContext(context);
+      const workspaces = await api.workspaces();
```
waxlamp, line 57: Probably need to catch errors here.

src/store/index.ts
```diff
@@ -0,0 +1,87 @@
...
+      commit.setWorkspaces(workspaces);
+    },
+    async fetchWorkspace(context, workspace: string) {
+      const { commit } = rootActionContext(context);
+      commit.setCurrentWorkspace({ name: workspace, nodeTables: [], edgeTables: [], graphs: [] });
```
waxlamp, line 62: We should maybe have a mutation that <i>un</i>sets workspaces, and use that here right?

src/store/index.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      const nodeTables = await api.tables(workspace, { type: 'node' });
+      const edgeTables = await api.tables(workspace, { type: 'edge' });
+      const graphs = await api.graphs(workspace);
```
waxlamp, line 66: Handle errors.

src/store/index.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      commit.setCurrentWorkspace({ name: workspace, nodeTables, edgeTables, graphs });
+    },
+  },
+});
+
+
```
waxlamp, line 73:
```suggestion

```

src/store/index.ts
```diff
@@ -0,0 +1,87 @@
...
+  rootGetterContext,
+  moduleGetterContext,
+};
+
+export type AppStore = typeof store;
+declare module 'vuex' {
+  interface Store<S> {
+    direct: AppStore;
+  }
+}
```
waxlamp, line 87: What does this do?

src/types.ts
```diff
@@ -5,6 +5,13 @@ export interface App {
   url: string;
 }
 
+export interface WorkspaceSpec {
+  name: string;
+  nodeTables?: string[];
+  edgeTables?: string[];
+  graphs?: string[];
+}
+
```
waxlamp, line 14: This seems like it should go in the state definition module. This "spec" is actually a spec for the workspace state data right?

Also, why are `nodeTables`, `edgeTables`, and `graphs` optional? They should be required, defaulting to empty arrays, no?

src/views/WorkspaceDetail.vue
```diff
@@ -139,15 +139,15 @@ export default Vue.extend({
...
-      graphs: [] as string[],
       requestError: null as string | null,
       loading: false,
     };
   },
 
   computed: {
+    nodeTables: () => store.getters.nodeTables,
+    edgeTables: () => store.getters.edgeTables,
+    graphs: () => store.getters.graphs,
```
waxlamp, line 150: Is it possible to use `mapGetters()`? Or does direct-vuex interfere with that?

src/views/WorkspaceDetail.vue
```diff
@@ -211,31 +211,14 @@ export default Vue.extend({
...
-        if (err.status === 404 && err.statusText === 'Workspace Not Found') {
-          this.$router.replace({name: 'home'});
-        } else {
-          throw err;
-        }
+      this.loading = true;
 
-        return;
+      await store.dispatch.fetchWorkspace(this.workspace);
+      if (Object.keys(this.$refs).length) {
```
waxlamp, line 217: Why is this guard needed?

src/views/WorkspaceDetail.vue
```diff
@@ -211,31 +211,14 @@ export default Vue.extend({
...
-        this.loading = false;
-        if (err.status === 404 && err.statusText === 'Workspace Not Found') {
-          this.$router.replace({name: 'home'});
-        } else {
-          throw err;
-        }
+      this.loading = true;
 
-        return;
+      await store.dispatch.fetchWorkspace(this.workspace);
```
waxlamp, line 216: Hoo boy, that's nice!

src/store/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    currentWorkspace: null,
+  } as State,
+  getters: {
+    nodeTables(state: State): string[] {
+      if (state.currentWorkspace !== null && state.currentWorkspace.nodeTables) {
+        return state.currentWorkspace.nodeTables;
+      }
+      return [];
+    },
+    edgeTables(state: State): string[] {
```
waxlamp, line 33: I want to get us in the habit of placing blank lines between multiline code blobs inside of object definitions. It aids readability immensely. This block of code is a great example of where it would help.

src/views/WorkspaceDetail.vue
AlmightyYakob: This is used in place of that, as mapGetters doesn't give any type information, and required workarounds all over the place if used.

waxlamp: Gotcha, I thought maybe direct-vuex had some way around that. Oh well.

src/store/index.ts
AlmightyYakob: This allows for typing info in the inject `$store` variable. The [reference](https://github.com/paleo/direct-vuex#create-the-store) from `direct-vuex`. I'm not sure if we'll want to use that injected variable, but I think we should keep the option there just in case, as that's how vuex works normally.

If we do keep this I'll be sure to add a comment though.

src/types.ts
AlmightyYakob: > This seems like it should go in the state definition module. This "spec" is actually a spec for the workspace state data right?

Yeah, I suppose a more appropriate name would be `WorkspaceState` or something, as it really just represents the state of the current workspace.

> Also, why are `nodeTables`, `edgeTables`, and `graphs` optional? They should be required, defaulting to empty arrays, no?

I don't think there's a good reason, I can make them required.

src/views/WorkspaceDetail.vue
AlmightyYakob: I think because we're calling update on the `created` lifecycle hook, this can run before the refs are actually available.

src/store/index.ts
AlmightyYakob: Good idea, I'll add that.

src/store/index.ts
AlmightyYakob: Should we display these errors to the user? Or just log them? I think I left it this way because if this call fails, the app as a whole won't work either, and so we'd want to see the error.

src/store/index.ts
AlmightyYakob: 4346a51

src/types.ts
AlmightyYakob: 0455145

src/store/index.ts
AlmightyYakob: d27a0dc

src/store/index.ts
AlmightyYakob: 4c837ce

src/store/index.ts
waxlamp: That's a good point. Eventually we'll want to use these failures to signal a general "health" indicator somewhere in the side or top bar of the app. Maybe that would be the time to actually catch errors here (and of course, such a mechanism would probably draw its health data from the store).

So, leave it alone for now. I've opened #70 to track the general idea.

src/store/index.ts
waxlamp: Same comment as above. Leave this as is for now.

src/store/index.ts
waxlamp: Great, thanks for the explanation.

Would you mind changing it to `ApplicationStore` or something? `AppStore` is going to cause us verbal confusion every time we say it out loud.

src/views/WorkspaceDetail.vue
waxlamp: Ok.

src/store/index.ts
```diff
@@ -47,19 +55,30 @@ const {
     setWorkspaces(state, workspaces: string[]) {
       state.workspaces = workspaces;
     },
-    setCurrentWorkspace(state, workspace: WorkspaceSpec) {
+
+    setCurrentWorkspace(state, workspace: WorkspaceState) {
       state.currentWorkspace = workspace;
     },
+
+    setEmptyWorkspace(state, name: string) {
```
waxlamp, line 63: This would be better named as `unsetWorkspace` IMO.

waxlamp: And there shouldn't be a `name` argument, right? Isn't the point of this to set the workspace to `null`? Or did I miss something?

src/store/index.ts
AlmightyYakob: 46b2dbf

src/store/index.ts
AlmightyYakob: You're right, I was under the impression that we needed the `name` property on the workspace to correctly display it, but it's tied to the route. I'll make that change. 

Also if it wasn't clear, the point of this mutation is to have it unset when switching between workspaces, so that the tables/networks of one don't bleed over into the other when making the transition.

AlmightyYakob: See 3af1a34, also I know you advocated for `unsetWorkspace`, but since the reciprocal call is `setCurrentWorkspace`, I changed it to `unsetCurrentWorkspace`

PR #69, opened 2020-04-01
Fix deletion api calls to order properly

AlmightyYakob:
This fixes a bug I just found/had encountered when doing the vuex stuff.

Before this change, this.$emit('deleted') would run right after
the forEach loop had finished kicking off asynchronous calls,
before they had finished. This lead to the fetching of new data
to happen in the middle of these deletion calls, which returned
incorrect/incomplete results.

### DIFF
--- a/src/components/DeleteGraphDialog.vue
+++ b/src/components/DeleteGraphDialog.vue
@@ -122,9 +122,7 @@ export default Vue.extend({
         workspace,
       } = this;
 
-      selection.forEach(async (graph) => {
-        await api.deleteGraph(workspace, graph);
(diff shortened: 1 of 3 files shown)


JackWilb:
Looks good. We might want to wait until we merge #66 and make sure we cover each instance

AlmightyYakob:
> We might want to wait until we merge #66 and make sure we cover each instance

Agree

AlmightyYakob:
> We might want to wait until we merge #66 and make sure we cover each instance

Looks like we're good with respect to this.

PR #71, opened 2020-04-02
Replace tslint with eslint, add rulesets

### LINKED ISSUE #33: Add pre-commit hook for client linting

### LINKED ISSUE #58: TSLint doesn't catch unused imports
I'm not sure if it doesn't catch other warnings, or what they may be, but this is the use case that I stumbled upon.

AlmightyYakob:
Fixes #58.
Closes #33.

Most of these changed files are fixes to the tooling, with `DownloadDialog.vue` as an example of what the linter might change/suggest changing. 

0eebea2 is an example of what the `yarn run fix` command would change on its own (besides [this](https://github.com/multinet-app/multinet-client/commit/0eebea2e3b8b362ed9cd9713dbd2b708e96b3da2#diff-a1633e2607538adfd60ef208e3682ac2R7) change, as I did that myself).

34a91d8 is an example of a more complex warning/error pointed out by the linter, which can't be automatically fixed, but which did improve the codebase as this change takes advantage of the parallel asynchronous calls.


AlmightyYakob:
Also forgot to make this a draft PR, but the goal of this is to get your opinions on the change, and then run these changes on the remainder of the codebase.

JackWilb:
These changes look reasonable to me. I'd love to hear any critiques that Roni has.

P.S. Also, I noticed a typo (not related to your code change), `downloadEnpoint` should be `downloadEndpoint`

waxlamp:
@AlmightyYakob, thanks for doing the work of setting this up, and thanks for letting me hack on it after the basics were working.

I did this as a huge stream of commits on purpose. The first three commits in my stream make relatively obvious changes (including the autofix changes from the linter itself). From there, I began layering on more difficult rule changes and fixes (things that required a careful handling of the code so as not to disturb its semantics). Some of these fixes did in fact impact how the code works, so later commits were added in order to fix those problems. This is why the later commits also get a bit more stormy.

After resolving all the lint issues, I had to tackle typescript issues. Several commits deal with that (including, predictably, typescript stuff I broke in the course of fixing lint errors). And finally, the cherry on the sundae was resolving some sticky issues that popped up at runtime.

CI is failing because the linting is not passing on the test script, but I will take care of that as soon as I can. For now if you could review these changes, and stress the netlify deployment (or a local build), that would be great. (I am not sure why the netlify deployment succeeded when CI didn't...).

JackWilb:
>I am not sure why the netlify deployment succeeded when CI didn't...
Netlify is probably not running the linter before it builds so there no failure. I'm guessing that the build works just fine

waxlamp:
> > I am not sure why the netlify deployment succeeded when CI didn't...
> 
> Netlify is probably not running the linter before it builds so there no failure. I'm guessing that the build works just fine

Sounds right, though I thought i'd seen in other projects where the previous steps in CI would prevent later ones from running. No big deal really, I'd actually prefer if Netlify succeeds even when the CI phase doesn't.

waxlamp:
Ok, this is ready for review!

@AlmightyYakob, does this PR also close #33?

AlmightyYakob:
> @AlmightyYakob, does this PR also close #33?

I'd say so, I mean technically we already had linting in place, but now it's truly what we wanted.

waxlamp:
> > @AlmightyYakob, does this PR also close #33?
> 
> I'd say so, I mean technically we already had linting in place, but now it's truly what we wanted.

Right, but I guess I'm asking if the pre-commit hook is set up to do what we need in light of the more comprehensive linting in this PR.

package.json
```diff
@@ -5,16 +5,18 @@
...
+    "lint:test:fix": "eslint --fix test/**/*.test.js",
+    "test": "tape -r esm test/**/*.test.js | tap-spec",
     "test:client:down": "sh ./scripts/stop.sh",
     "test:client:restart": "yarn test:client:down; yarn test:client:up",
-    "test": "tape -r esm test/**/*.test.js | tap-spec"
+    "test:client:up": "sh ./scripts/start.sh",
+    "test:server:clean": "cd .. && PIPENV_DONT_LOAD_ENV=1 pipenv run test-server-clean",
+    "test:server:down": "cd .. && PIPENV_DONT_LOAD_ENV=1 pipenv run test-server-down",
+    "test:server:restart": "yarn test:server:down; yarn test:server:clean && yarn test:server:up",
+    "test:server:up": "cd .. && PIPENV_DONT_LOAD_ENV=1 pipenv run test-server-up"
```
JackWilb, line 19: These will need to be re-thought at some time. I'm happy to defer them to later, though

src/views/GraphDetail.vue
```diff
@@ -378,6 +419,7 @@ export default Vue.extend({
 
       this.nodeTypes = graph.nodeTables;
       this.edgeTypes = [graph.edgeTable];
+      // eslint-disable-next-line no-underscore-dangle
       this.nodes = nodes.nodes.map((d) => d._id);
```
AlmightyYakob, line 423: FWIW, eslint allows you to [provide exceptions](https://eslint.org/docs/rules/no-underscore-dangle#allow) to this rule, so we could add `_id` to that if we wanted.

src/components/AboutDialog.vue
```diff
@@ -1,49 +1,62 @@
...
         class="px-4 pt-4 pb-1"
-        >
+      >
         Multinet is a system for storing and processing <a-ext
-          href="https://vdl.sci.utah.edu/mvnv/">multivariate networks</a-ext>. Learn more and explore the code
-        at <a-ext href="https://github.com/multinet-app/multinet">GitHub</a-ext>.
+          href="https://vdl.sci.utah.edu/mvnv/"
+        >
+          multivariate networks
+        </a-ext>. Learn more and explore the code
```
JackWilb, line 35: Looks like formatting the `a-ext` tags like this adds a space to the end of the links

src/components/WorkspaceDialog.vue
```diff
@@ -8,7 +8,7 @@
     origin="center center"
     transition="scale-transition"
   >
-    <template>
+    <template v-slot:activator>
```
AlmightyYakob: Commenting more on this commit description than the line itself:

The reason for the error after removing the `v-slot` identifier altogether, is because it needs to know that it's a v-slot, specifically the `activator` slot, as the "name" of this slot is actually `activator`. When you specify `v-slot:activator="{ on }"`, you're specifying what parameters you want to access, from the parameters that are exposed from the slot definition. Here is the list of the slot arguments, which you can find on the Vuetify [menu](https://vuetifyjs.com/en/components/menus/) component page:

```
{
  attrs: { role: string, aria-haspopup: boolean, aria-expanded: string }
  on: { [eventName]: eventHandler }
  value: boolean
}
```

But since we weren't actually using the parameter that we exposed, it caused a linting error.

src/App.vue
```diff
@@ -14,7 +14,7 @@
 import Sidebar from '@/components/Sidebar.vue';
 import Vue from 'vue';
 
-import apps from './appregistry.json';
+const apps = require('./appregistry.json');
```
AlmightyYakob, line 18: Is this being done to satisfy the linter complaining about not specifying `resolveJsonModule`? If so, the following addition to `tsconfig.json` will fix this:

```
  "compilerOptions": {
    ...
    "resolveJsonModule": true,
```

Which I've tested locally and does work.

src/views/TableDetail.vue
```diff
@@ -13,34 +16,41 @@
...
 
-          <v-divider></v-divider>
+          <v-divider />
 
           <v-list-item
+            v-for="t in tables"
+            :key="t"
             ripple
-            v-for="table in tables"
-            :key="table"
```
JackWilb, line 21: This change makes the code slightly less readable on first glance

src/views/TableDetail.vue
```diff
@@ -141,6 +164,18 @@ export default Vue.extend({
...
+    workspace() {
+      this.update();
+    },
+    table() {
+      this.update();
+    },
+
+    pagination() {
+      this.update();
+    },
```
JackWilb, line 177: We can remove this whitespace

src/views/WorkspaceDetail.vue
```diff
@@ -87,16 +107,23 @@
           </v-card>
         </v-flex>
 
-        <v-flex md6 px-5 py-3>
-          <v-card color="transparent" flat text>
+        <v-flex
+          md6
+          px-5
+          py-3
+        >
```
JackWilb, line 114:
```suggestion
        <v-flex
          class="md6 px-5 py-3"
        >
```

src/views/WorkspaceDetail.vue
```diff
@@ -75,8 +87,16 @@
       <v-layout
         wrap
       >
-        <v-flex md6 px-5 py-3>
-          <v-card color="transparent" flat text>
+        <v-flex
+          md6
+          px-5
+          py-3
+        >
```
JackWilb, line 94:
```suggestion
        <v-flex
          class="md6 px-5 py-3"
        >
```

src/components/WorkspaceDialog.vue
```diff
@@ -8,7 +8,7 @@
     origin="center center"
     transition="scale-transition"
   >
-    <template v-slot:activator>
+    <template v-slot:activator="{ }">
```
AlmightyYakob, line 11: The reason I believe this is happening, is because currently the whole dialog is wrapped in a `v-menu`, which is providing the "Getting Started" popover. This should probably go into `Sidebar.vue` or somewhere more appropriate, but regardless probably shouldn't be wrapping whole components like this. I say we fix this in a follow-up issue.

src/views/NodeDetail.vue
```diff
@@ -82,23 +83,26 @@
                     class="pt-0"
                     color="transparent"
                   >
-                    <v-container fluid pa-0>
+                    <v-container
+                      fluid
+                      pa-0
+                    >
```
JackWilb, line 89:
```suggestion
                    <v-container
                      fluid
                      class="pa-0"
                    >
```

src/views/TableDetail.vue
AlmightyYakob: I agree, and the linter seems fine with the order of:

```
v-for
:key
:to
ripple
```

So my vote is for that

JackWilb: I meant using t instead of table here. I'm okay with any order, really 😄

AlmightyYakob:
> Right, but I guess I'm asking if the pre-commit hook is set up to do what we need in light of the more comprehensive linting in this PR.

I believe so, since it's still just running `vue-cli-service lint` on any `vue` or `ts` files.

src/views/GraphDetail.vue
waxlamp: Yeah but this is literally the one time it comes up so I thought a one-off silencing was simpler.

src/App.vue
waxlamp: I don't quite remember, but I think it had to do with the fact that we are really loading `appregistry.json` at runtime, hence the need for `require()`. Or maybe I just misunderstood what the errors were. In any case, we will be retooling this mechanism very soon in order to get Netlify to build properly linked DNSed vis apps, so I'm ok with leaving this as is and not inflating the ruleset on its account.

src/components/WorkspaceDialog.vue
waxlamp: Can you file an issue and coordinate with @jtomeck? I don't seem to understand what's going wrong here as well as you do.

package.json
waxlamp: Can you file an issue with a description of what needs to change? This particular change was simply the linter alphabetizing the scripts in package.json, so this wouldn't be the right time to make other, non-lint-related changes to this.

src/views/TableDetail.vue
waxlamp: I had to use something besides `table` to avoid shadowing the `table` variable declared in the containing scope. Iteration using a pattern like `t in tables` is pretty idiomatic so let's leave it.

src/views/TableDetail.vue
waxlamp: I want to include whitespace like this between method declarations for readability. I made the change here in the interest of starting a new habit. But as future work occurs I plan to introduce this type of whitespace in the other modules.

My thinking is just to have surrounding blank lines for multi-line functions in the component declarations. If you have a series of simple one-liners, you could choose to group those in a block, but in general I'd say the extra whitespace drastically improves legibility.

src/views/WorkspaceDetail.vue
waxlamp: The `class` keyword isn't needed and reduces readability in this case. I can remove the linebreaks though.

src/views/WorkspaceDetail.vue
waxlamp: Same as above.

src/views/NodeDetail.vue
waxlamp: Same as above.

waxlamp:
> > Right, but I guess I'm asking if the pre-commit hook is set up to do what we need in light of the more comprehensive linting in this PR.
> 
> I believe so, since it's still just running `vue-cli-service lint` on any `vue` or `ts` files.

I added a "closes" clause to the PR description.

src/views/WorkspaceDetail.vue
JackWilb: This is the only place that it's formatted like this though, right? Aren't all of the others as classes?

waxlamp: Turns out the linter doesn't want that many attributes on one line, so I'll leave this as is.

src/views/TableDetail.vue
JackWilb: It just wasn't consistent with the other functions here. I agree with your comment on readability though

package.json
JackWilb: See #73

src/views/NodeDetail.vue
AlmightyYakob: This seems to work, which makes the syntax a bit nicer for us, but I can't seem to find this documented anywhere. Maybe I'm just missing it in their docs, but all of their examples use the `class="..."` syntax. For that reason, I think it's better to stick to the recommended way of doing it, instead of using this non-standard way. I'm okay leaving this PR as is, just thought I'd throw my two cents in. I'd also be curious if you had a link to documentation that references this syntax @waxlamp.

src/components/AboutDialog.vue
waxlamp: See c22cfa9425286eea7299afcb13bfc43ee53a970e.

AlmightyYakob:
@waxlamp LGTM :+1:

PR #72, opened 2020-04-02
Fix downloaders and lingering selection

AlmightyYakob:
This probably just slipped through the cracks when we split up `ItemPanel`. If you'll notice, before this change, downloading a selection of tables/networks would keep the selection afterwards. 

Also, if you were to navigate to a different workspace after making a selection, that same selection would still be active on the second workspace, and would show up in the download/delete dialog.

This PR addresses both of these issues.

### DIFF
--- a/src/components/DownloadDialog.vue
+++ b/src/components/DownloadDialog.vue
@@ -170,7 +170,7 @@ export default Vue.extend({
         URL.revokeObjectURL(link.href);
       }
 
-      this.$emit('downloaded');
+      this.$emit('downloaded', selection);
       this.loading = false;
       this.dialog = false;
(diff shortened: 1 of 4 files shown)


AlmightyYakob:
> I didn't try this out but the code changes look fine, and I trust that you verified the correct behavior.

:+1: I did test it locally, but the Netlify preview would also show the new behavior.

waxlamp:
> > I didn't try this out but the code changes look fine, and I trust that you verified the correct behavior.
> 
> +1 I did test it locally, but the Netlify preview would also show the new behavior.

If you verify it on netlify, then you can go ahead and merge.

PR #86, opened 2020-05-08
Use appregistry.json for the quick links

JackWilb:
Reduces the number of places where we manually define the link to the nodelink and adjacency matrix applications

### DIFF
--- a/src/views/FrontPage.vue
+++ b/src/views/FrontPage.vue
@@ -156,7 +156,7 @@ export default Vue.extend({
           // eslint-disable-next-line global-require
           image: require('../assets/placard/miserables.jpg'),
           text: 'Explore the Les Miserables dataset using an interactive adjacency matrix. See the factions and relationships for yourself!',
-          href: `${this.apps.filter((d) => d.name === 'Adjacency Matrix')[0].url}?workspace=miserables&graph=miserables`,
+          href: `${this.apps.filter((d) => d.name === 'Adjacency Matrix')[0].url}/?workspace=miserables&graph=miserables`,
         },
         {
           title: 'Les Miserables - Node Link Diagram',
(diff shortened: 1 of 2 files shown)


src/views/FrontPage.vue
```diff
@@ -112,9 +112,17 @@
...
+import Vue, { PropType } from 'vue';
+import { App } from '@/types';
 
 export default Vue.extend({
+  props: {
+    apps: {
+      type: Array as PropType<App[]>,
+      required: true,
+    },
+  },
```
AlmightyYakob, line 124: Is the `apps` prop passed in here? https://github.com/multinet-app/multinet-client/blob/b36a8b9ba20e89498c9daf31ba3e887bff558890/src/App.vue#L6-L9

If that's the case, I don't think we should pass `apps` as a prop to every page via `router-view`, only some of which use it. Since it just imports `appregistry.json`, we should just directly import that in whichever files need it. I know you didn't write that, but just something I noticed. I'd be happy to make a separate issue for it if we want.

JackWilb: I believe that's where it's picking it up from, yes. I was following the same usage of the apps prop in the `GraphDetail.vue` file so it will need updating there, too

waxlamp: In fact, this seems like a good time to use the application store. The appregistry is configured via a JSON file, but it doesn't change once the application starts up. So it seems like it makes sense to inject it into the store once at the beginning, and always map it from the store into any component that needs it.

AlmightyYakob: I'd say the fact that it doesn't change is actually a reason against putting it in the store, as if it's never changing there's no benefit to doing that over simply importing the file, and would add unnecessary complexity. I'll make a separate issue for this though.

waxlamp: https://github.com/multinet-app/multinet-client/issues/87

PR #88, opened 2020-05-11
Remove lingering TODOs and update calls

### LINKED ISSUE #65: Remove use of update events when vuex is added
Follow up from #59

https://github.com/multinet-app/multinet-client/blob/68df20d2011dffd017b131512598a2c511ab5661/src/views/WorkspaceDetail.vue#L260-L261

https://github.com/multinet-app/multinet-client/blob/68df20d2011dffd017b131512598a2c511ab5661/src/App.vue#L3-L4

https://github.com/multinet-app/multinet-client/blob/68df20d2011dffd017b131512598a2c511ab5661/src/App.vue#L5-L6

https://github.com/multinet-app/multinet-client/blob/68df20d2011dffd017b131512598a2c511ab5661/src/App.vue#L27-L33

AlmightyYakob:
Fixes #65.

### DIFF
--- a/src/components/NetworkPanel.vue
+++ b/src/components/NetworkPanel.vue
@@ -120,7 +120,6 @@ import DownloadDialog from '@/components/DownloadDialog.vue';
 
 import store from '@/store';
 
-
 export default Vue.extend({
   name: 'GraphPanel',
 
--- a/src/components/NetworkPanel.vue
+++ b/src/components/NetworkPanel.vue
@@ -118,6 +118,9 @@ import DeleteGraphDialog from '@/components/DeleteGraphDialog.vue';
 import GraphDialog from '@/components/GraphDialog.vue';
 import DownloadDialog from '@/components/DownloadDialog.vue';
 
+import store from '@/store';
+
+
 export default Vue.extend({
   name: 'GraphPanel',
 
@@ -217,7 +220,7 @@ export default Vue.extend({
       });
     },
 
-    cleanup(selection?: string[]) {
+    async cleanup(selection?: string[]) {
       this.singleSelected = null;
 
       if (selection) {
@@ -226,7 +229,8 @@ export default Vue.extend({
         });
       }
 
-      this.$emit('update');
+      await store.dispatch.fetchWorkspace(this.workspace);
+      this.clearCheckboxes();
     },
   },
 });
--- a/src/views/WorkspaceDetail.vue
+++ b/src/views/WorkspaceDetail.vue
@@ -102,7 +102,6 @@
               :items="graphs"
               :node-tables="nodeTables"
               :edge-tables="edgeTables"
-              @update="update"
             />
           </v-card>
         </v-flex>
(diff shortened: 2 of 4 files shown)


src/components/NetworkPanel.vue
```diff
@@ -118,6 +118,9 @@ import DeleteGraphDialog from '@/components/DeleteGraphDialog.vue';
 import GraphDialog from '@/components/GraphDialog.vue';
 import DownloadDialog from '@/components/DownloadDialog.vue';
 
+import store from '@/store';
+
+
```
waxlamp, line 123:
```suggestion
```

src/views/WorkspaceDetail.vue
```diff
@@ -222,8 +220,7 @@ export default Vue.extend({
           this.editing = false;
           this.requestError = null;
 
-          // TODO: REMOVE THIS REF WHEN VUEX IS ADDED
-          this.$emit('update');
+          store.dispatch.fetchWorkspaces();
```
waxlamp, line 223: Does this need an `await` plus possibly some other visual state logic (like the `this.loading` prop in the hunk below)? Could that be related to the bug @JackWilb mentioned?

AlmightyYakob:
> There's a new bug introduced here that happens under quite strict circumstance:
> 
> 1. Click on a workspace to load it in
> 2. As it's loading, click on another workspace.
> 3. The UI shows information from both place. The networks and tables are from the first workspace clicked and the name of the workspace (in the top bar) is from the second workspace clicked
> 
> I would guess that this is some anomaly of the async await logic that was added

I don't really experience a huge problem with this. I can kind of see the issue you're talking about. If I follow the steps you outlined, the networks and tables don't match the workspace name for a split second. However, after that, the data updates to match the workspace name. Does it persist permanently for you?

JackWilb:
>Does it persist permanently for you?

It does, but only in a specific circumstance. It's when the 2nd workspace tables + graphs load before the first. So when the first workspace is much larger, it will persist in the wrong state. It actually looks like this behavior exists in the current deployed app as well

Here is a screenshot with the issue. The miserables workspace is loaded, but the tables from evo-bio persist
Screen Shot 2020-05-18 at 2 05 32 PM

waxlamp:
@JackWilb, since this bug didn't show up in this branch, can you file an issue describing it in detail and we'll figure out a solution independently?

AlmightyYakob:
@JackWilb Thank you for the clarification, that makes perfect sense. I think that would be better left as a separate issue like @waxlamp said. I think I have a path forward to fix this issue.

JackWilb:
> @JackWilb Thank you for the clarification, that makes perfect sense. I think that would be better left as a separate issue like @waxlamp said. I think I have a path forward to fix this issue.

Opened #89

src/views/WorkspaceDetail.vue
AlmightyYakob: I don't think this is related to the bug @JackWilb found, as this is in the `renameWorkspace` method. Awaiting this should have no effect, as it's the last call in the method.

PR #95, opened 2020-05-29
Add netlify.toml

AlmightyYakob:
Adds file configuration for netlify, and adds the ability to specify different environment variables for production builds vs branch previews

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,6 +2,9 @@
 publish = "dist"
 command = "yarn build"
 
+  [build.environment]
+  VUE_APP_GA_TAG = "UA-157426670-1"
+
 [context.deploy-preview.environment]
 VUE_APP_MULTINET_HOST = "https://multinet-app-testing.herokuapp.com"
 
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,9 @@
+[build]
+  publish = "dist"
+  command = "yarn build"
+
+[context.deploy-preview.environment]
+  VUE_APP_MULTINET_HOST = "https://multinet-app-testing.herokuapp.com"
+
+[context.production.environment]
+  VUE_APP_MULTINET_HOST = "https://api.multinet.app"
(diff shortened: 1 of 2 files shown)


AlmightyYakob:
Draft PR until it's been tested out in some manner

AlmightyYakob:
Seems to be working, so I've marked it as ready for review

netlify.toml
```diff
@@ -0,0 +1,12 @@
+[build]
+  publish = "dist"
+  command = "yarn build"
+
+  [build.environment]
+    VUE_APP_GA_TAG = "UA-157426670-1"
```
waxlamp, line 6: I don't think we should keep this particular setting in version control.

netlify.toml
```diff
@@ -0,0 +1,12 @@
+[build]
+  publish = "dist"
```
waxlamp, line 2: I think standard TOML indentation is to keep the key/value lines flush with the header (even when a whole section block is indented, as you have below).

netlify.toml
AlmightyYakob: I'm okay with removing it, but FWIW, we shouldn't have any expectation of privacy regarding this value, as it's totally observable on the client. Also, google will reject requests with this tag that aren't from the whitelist we've configured in the console.

waxlamp: So, there's no way someone could accidentally send their own analytics data to our data sink? That's more what I'm concerned about, rather than privacy. If that's the case, we can leave this as-is.

AlmightyYakob: If configured right I believe so, however I'm not sure if it currently is, or if I have the permissions to ensure that, so for now I'll remove it from here.

netlify.toml
AlmightyYakob: da6f456

netlify.toml
waxlamp: Oh I don't think you need to remove it. We just need to make sure that we have GA set up the right way. I naively assumed that by default, this kind of "leak" wouldn't be protected against, but any thought at all shows that Google must have built in this protection, for fear of making their analytics product completely useless to people.

PR #104, opened 2020-06-25
Refresh visible workspaces and data on logout

### LINKED ISSUE #103: Refresh the workspace list on logout
Since, in general, a logged in user can see more workspaces than a logged out user, it's critical to force a refetch of the workspace list on log out.

This may also entail changing the active workspace that was being viewed at the moment of log out (perhaps simply redirecting the user to the home page).

On log in, no special action is needed because of how Google OAuth works: after the login action is completed, the system redirects the user to a chosen URL, which in our case always "forces a refresh" of the workspace list (really, it's just loading it for the first time, for the browser session that starts after redirect).

waxlamp:
Closes #103

### DIFF
--- a/src/components/LoginMenu.vue
+++ b/src/components/LoginMenu.vue
@@ -111,7 +111,11 @@ export default {
       // This is to prevent the logged-out user from continuing to look at, e.g.,
       // workspaces or tables they may have been viewing at the time of logout.
       await store.dispatch.logout();
-      this.$router.push('home');
+
+      // Avoid illegal duplicate navigation if we are already on the Home view.
+      if (this.$router.currentRoute.name !== 'home') {
+        this.$router.push('home');
+      }
     },
   },
 };
--- a/src/components/LoginMenu.vue
+++ b/src/components/LoginMenu.vue
@@ -105,7 +105,14 @@ export default {
   },
 
   methods: {
-    logout: () => store.dispatch.logout(),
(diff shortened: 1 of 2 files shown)


src/components/LoginMenu.vue
```diff
@@ -105,7 +105,14 @@ export default {
...
 
   methods: {
-    logout: () => store.dispatch.logout(),
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    async logout(this: any) {
+      // Perform the logout action, then redirect the user to the home page.
+      // This is to prevent the logged-out user from continuing to look at, e.g.,
+      // workspaces or tables they may have been viewing at the time of logout.
+      await store.dispatch.logout();
+      this.$router.push('home');
```
AlmightyYakob, line 114: If you're already on the home page, this causes an error in the console with `vue-router`, because you're trying to navigate to the route that you're already at. I'd think refreshing the page here would be a good idea, to both ensure that new data is loaded, and that old data doesn't persist. I believe `this.$router.go()` would accomplish this.

waxlamp: I'd prefer not to refresh at this point, since it would feel disruptive. I will look for a way to avoid the console error.

waxlamp: See 3d82367686ab8bae95335afba8f39ef1473fd5b3.

PR #105, opened 2020-06-26
Disable "new workspace" button when user is not logged in

waxlamp:
This PR also removes the "intro" popover, as (1) it has become sort of annoying and (2) its advice no longer applies.

Since the way to really get started with this app is to create an account and log in, I'd like to find a UI design principle that makes the login mechanism more obvious and inviting (currently it's hidden behind a general "user account" badge).

### DIFF
--- a/src/components/WorkspaceDialog.vue
+++ b/src/components/WorkspaceDialog.vue
@@ -79,26 +79,6 @@
         </v-card>
       </v-dialog>
     </template>
-    <v-card>
-      <v-card-title class="get-started-title pb-2">
-        Getting Started
-      </v-card-title>
-      <v-card-text class="pb-5">
-        Click <strong>NEW WORKSPACE</strong> to create a workspace or select an existing one from the Workpaces list.
-      </v-card-text>
-      <v-divider />
-      <v-card-actions>
-        <v-spacer />
-        <v-btn
-          id="got-it"
-          color="primary"
-          small
-          @click="popover = false"
-        >
-          Got it!
-        </v-btn>
-      </v-card-actions>
-    </v-card>
   </v-menu>
 </template>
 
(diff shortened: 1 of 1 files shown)


waxlamp:
> There appears to be a new bug in this PR. When I click create workspace after refreshing (and on logging in, since it redirects to the home page) the modal appears for a split second and disappears. It's only persistent and usable on a second click. Let me know if the same behavior is happening for you.
> 
> I verified that this happens on chromium and firefox for me on my linux box, and on chrome on my mac.

Can you paste in any console errors that appear when this happens?

JackWilb:
> Can you paste in any console errors that appear when this happens?

There are no errors in the console, but I think I've tracked down the issue to the v-menu having the `v-model="popover"` set on it. Removing it seems to fix the error.

waxlamp:
> > Can you paste in any console errors that appear when this happens?
> 
> There are no errors in the console, but I think I've tracked down the issue to the v-menu having the `v-model="popover"` set on it. Removing it seems to fix the error.

Thanks for tracking this down. I pushed cdc4d95112821dd7693e6569a45279a20272fc65 to fix it; can you verify that it's working now?

src/components/WorkspaceDialog.vue
```diff
@@ -15,12 +14,17 @@
         width="500"
       >
         <template v-slot:activator="{ on }">
+          <!-- See
+            https://github.com/vuetifyjs/vuetify/issues/4482#issuecomment-476689473
+            for why the `dark` prop is tied to the disabled state of the button.
+          -->
           <v-btn
             id="add-workspace"
+            :disabled="userInfo == null"
```
AlmightyYakob, line 23:
```suggestion
            :disabled="userInfo === null"
```

waxlamp: Thanks. I'm surprised the linter didn't catch this...

PR #122, opened 2020-09-09
Replace use of cookies for token storage with localStorage

AlmightyYakob:
Depends on https://github.com/multinet-app/multinet-server/pull/472
Depends on https://github.com/multinet-app/multinetjs/pull/19.

Notable changes:

* Move login logic from `App.vue` to `LoginMenu.vue`
* Add method that checks the URL for an included loginToken, handling it if present.
* Use `userInfo` and `logout` functions of `multinetjs`
* Add interface for storing/retrieving login tokens through localStorage
* Fix incorrect encoding of `return_url` when logging in

PR #124, opened 2020-09-16
Pass login tokens to the vis apps

### LINKED ISSUE #123: Pass token along to vis apps in querystring
Since we're now using tokens for the authentication, we need to pass them along to the vis apps so they can access private data.

JackWilb:
Closes #123

Adds logic to pass the login token from the main client to the vis apps via the querystring. 

I also had to update yarn.lock, since it wasn't updated in #122 after the version was bumped.

### DIFF
--- a/src/views/GraphDetail.vue
+++ b/src/views/GraphDetail.vue
@@ -93,7 +93,7 @@
                 v-for="app in apps"
                 :key="app.name"
                 class="pl-2"
-                :href="`${app.url}/?workspace=${workspace}&graph=${graph}${loginToken}`"
+                :href="`${app.url}/?workspace=${workspace}&graph=${graph}#${loginToken}`"
                 target="_blank"
               >
                 <v-list-item-avatar class="mr-3">
@@ -387,7 +387,7 @@ export default Vue.extend({
       const token = getLoginToken();
 
       if (token !== null) {
-        return `&loginToken=${token}`;
+        return `loginToken=${token}`;
       }
 
       return '';
--- a/src/views/GraphDetail.vue
+++ b/src/views/GraphDetail.vue
@@ -93,7 +93,7 @@
                 v-for="app in apps"
                 :key="app.name"
                 class="pl-2"
-                :href="`${app.url}/?workspace=${workspace}&graph=${graph}&token=${userToken}`"
+                :href="`${app.url}/?workspace=${workspace}&graph=${graph}${loginToken}`"
                 target="_blank"
               >
                 <v-list-item-avatar class="mr-3">
@@ -383,8 +383,14 @@ export default Vue.extend({
     drawerIcon(): string {
       return this.panelOpen ? 'expand_more' : 'expand_less';
     },
-    userToken() {
-      return getLoginToken();
+    loginToken(): string {
+      const token = getLoginToken();
+
+      if (token !== null) {
+        return `&loginToken=${token}`;
+      }
+
+      return '';
     },
   },
   watch: {
--- a/src/views/GraphDetail.vue
+++ b/src/views/GraphDetail.vue
@@ -93,7 +93,7 @@
                 v-for="app in apps"
                 :key="app.name"
                 class="pl-2"
-                :href="`${app.url}/?workspace=${workspace}&graph=${graph}#${loginToken}`"
+                :href="`${app.url}/?workspace=${workspace}&graph=${graph}${loginToken}`"
                 target="_blank"
               >
                 <v-list-item-avatar class="mr-3">
@@ -387,7 +387,7 @@ export default Vue.extend({
       const token = getLoginToken();
 
       if (token !== null) {
-        return `loginToken=${token}`;
+        return `#loginToken=${token}`;
       }
 
       return '';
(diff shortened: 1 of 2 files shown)


src/views/GraphDetail.vue
```diff
@@ -93,7 +93,7 @@
                 v-for="app in apps"
                 :key="app.name"
                 class="pl-2"
-                :href="`${app.url}/?workspace=${workspace}&graph=${graph}`"
+                :href="`${app.url}/?workspace=${workspace}&graph=${graph}&token=${userToken}`"
```
AlmightyYakob, line 96: I'd prefer if this was `loginToken`, so it's consistent with the other use of it.

```suggestion
                :href="`${app.url}/?workspace=${workspace}&graph=${graph}&loginToken=${userToken}`"
```

AlmightyYakob (changes requested):
Just one small suggestion, otherwise looks good

src/views/GraphDetail.vue
AlmightyYakob: Also, it might be good to have a computed function for the end `&loginToken=...`, which would return the empty string if the token isn't present. This would prevent always passing `&loginToken=` whenever the user isn't logged in.

src/views/GraphDetail.vue
```diff
@@ -382,6 +383,9 @@ export default Vue.extend({
     drawerIcon(): string {
       return this.panelOpen ? 'expand_more' : 'expand_less';
     },
+    userToken() {
```
waxlamp, line 386: Rename this to `loginToken()` to reflect @AlmightyYakob's comments above.

waxlamp (changes requested):
Thanks for taking initiative on this, @JackWilb. I agree with @AlmightyYakob's overall review/comments.

src/views/GraphDetail.vue
JackWilb: See 89f51e5

src/views/GraphDetail.vue
JackWilb: See 89f51e5

src/views/GraphDetail.vue
```diff
@@ -387,7 +387,7 @@ export default Vue.extend({
       const token = getLoginToken();
 
       if (token !== null) {
-        return `&loginToken=${token}`;
+        return `loginToken=${token}`;
```
waxlamp, line 390: I think you want to keep the `#` here; in case there is no token, line 96 above will have a `#` in the URL with nothing following it. Cleaner would be to keep the `#` here so that it won't appear in the URL at all if there's no token.

yarn.lock
```diff
@@ -5852,10 +5852,10 @@ multimatch@^2.1.0:
...
     minimatch "^3.0.0"
 
-multinet@0.16.0:
-  version "0.16.0"
-  resolved "https://registry.yarnpkg.com/multinet/-/multinet-0.16.0.tgz#7361fc87fbc84745f82f45bb8dd5ccd4d6a1f39b"
-  integrity sha512-aqbuRPG5XwGgK1ChhQwGmKkMJcrefbB7H0bKtLIU3OJp1iDwKSWhQ+wdfzPHvPweRt4f798suXdybN1v/uf8hw==
+multinet@0.17.0:
+  version "0.17.0"
+  resolved "https://registry.yarnpkg.com/multinet/-/multinet-0.17.0.tgz#05b30b85d29af8cb877bed502d8cacdabe8796ed"
+  integrity sha512-EC+ALY7ciwSmbkC49NE5A0OmFYxjt/QrTPQf7sXDeL6FUi74UKaammlN3E11T4Pur3kNx686AycmKSbgCkPHNA==
```
waxlamp, line 5858: A reminder: in general, please keep unrelated changes in their own PRs (even if it's a tiny change like this one).

waxlamp (changes requested):
Looks good overall. Just the one request.

src/views/GraphDetail.vue
JackWilb: I had moved it to the `url` param, above in this file, so that if we add another parameter to the fragment it would work as intended. Happy to make this change if you prefer

AlmightyYakob: Same concern that Roni raised. If we're concerned about supporting multiple fragments in the future, it'd probably be good to replace this computed function with a `fragment` computed function, which could aggregate any desired fragment sections, and return something like:

```
// Create this list somehow
const fragments = ['loginToken=abcd', ...];
return fragments.length ? `#${fragments.join("&")}` : '';
```

However, it's not something that really needs to be addressed in this PR, just a thought.

yarn.lock
JackWilb: The version is bumped in package.json in master, but not in yarn.lock. When I ran `yarn install`, it was automatically upgraded. Also, version 0.17.0 is required for all the new auth changes. Would you like me to extract this out into its own PR that this Pr depends on?

AlmightyYakob: In that case it looks like we already messed up. I don't think this is worth extracting into its own PR, since this should already be in master anyway.

src/views/GraphDetail.vue
JackWilb: Addressed in 5877967

yarn.lock
JackWilb: I'll leave it as is for now and in future make similar changes as a separate PR

AlmightyYakob (approved):

waxlamp (approved):

PR #125, opened 2020-10-01
Add smart column type detection and UI for setting those types

waxlamp:
This PR adds a long sought after feature: automatic type detection of CSV columns, and UI for updating them before uploading.

The way this works is to:
1. use `papaparse` to parse each line into a JS data object.
2. subject each value to a rudimentary type detector (see below).
3. use a simple heuristic to detect the closest type to suggest for each column
4. present a dropdown UI to let the user adjust the types before upload

The types currently recognized are:
- `label`: a string type that is meant to hold arbitrary data for creating textual labels in a visualization
- `category`: a type holding a limited number of distinct values, meant to drive visualization for an ordinal or categorical variable
- `number`: a quantitative value
- `date`: a value that looks like a date in some way

`label` is a fallback categorization; essentially any column that doesn't fall into the other types is assigned as a label. This works out naturally because CSV data, by definition, comes in the door as a set of strings. `label` is also suggested for fields named `_key`, `_from`, and `_to`, since those are special types in Multinet/ArangoDB, and need to actually be string values, even if they look like numbers, etc.

`category` is detected if the column contains 10 or fewer distinct values.

`number` is detected if every value in the column is coercible to JavaScript's `Number` type (using `!Number.isNaN(Number(x))` as the litmus test).

`date` is detected if every value in the column is coercible to DayJS's date wrapper type. However, `number` receives priority over date, because DayJS considers low numeric values to be valid dates, and in the majority of cases, such columns are actually intended as numbers.

All of the above is meant only for heuristic suggestion of the column types. The user gets final say by manipulating the resulting dropdowns.

This PR just offers a UI component for the feature; as things stand, the server would need to take the type information and do something with it prior to validation. The other option is to change how the CSV uploader works so that it accepts JSON lines instead of CSV, and use the type fields to properly convert the data before upload. This would still require the server to try to use the `datetime` module (or something more advanced, such as `arrow` or `pendulum` to attempt to convert the string values to dates, then settle them into the database in some canonical format, probably [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)).

TODO
- [x] `trim()` the `value` variable when testing for Numberness
- [x] properly clear out the dropdowns when the X is clicked in the file upload dialog

### DIFF
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -111,6 +111,7 @@
 <script lang="ts">
 import Vue, { PropType } from 'vue';
 import Papa, { ParseResult } from 'papaparse';
+import dayjs from 'dayjs';
 
 import api from '@/api';
 import { FileType } from '@/types';
@@ -230,6 +231,11 @@ export default Vue.extend({
             // convert to 0, so those are excluded specifically.
             if (value !== '' && !Number.isNaN(Number(value))) {
               entry.number += 1;
+            }
+
+            // See if the value looks like a date.
+            if (dayjs(value).isValid()) {
+              entry.date += 1;
             }
           });
         },
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -96,6 +96,7 @@
 
 <script lang="ts">
 import Vue, { PropType } from 'vue';
+import Papa, { ParseResult } from 'papaparse';
 
 import api from '@/api';
 import { FileType } from '@/types';
@@ -157,6 +158,50 @@ export default Vue.extend({
         this.fileName = this.fileName || getFileName(file);
         this.fileUploadError = null;
       }
+
+      interface TypeScore {
+        strings: Set<string>;
+        number: number;
+        date: number;
+        total: number;
+      }
+      const columnTypes = new Map<string, TypeScore>();
+
+      Papa.parse(file, {
+        header: true,
+        skipEmptyLines: true,
+        step(row: ParseResult<{}>) {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          const data = row.data as { [key: string]: any };
+
+          Object.keys(data).forEach((key: string) => {
+            if (!columnTypes.has(key)) {
+              columnTypes.set(key, {
+                strings: new Set(),
+                number: 0,
+                date: 0,
+                total: 0,
+              });
+            }
+
+            const entry = columnTypes.get(key);
+            if (entry === undefined) {
+              throw new Error('impossible');
+            }
+
+            const value = data[key];
+            entry.total += 1;
+            entry.strings.add(value);
+            if (value !== '' && !Number.isNaN(Number(value))) {
+              entry.number += 1;
+            }
+          });
+        },
+
+        complete() {
+          console.log(columnTypes);
+        },
+      });
     },
 
     handleUploadProgress(evt: { loaded: number; total: number; [key: string]: unknown }) {
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -72,6 +72,20 @@
               />
             </v-col>
           </v-row>
+
+          <v-row>
+            <v-col
+              v-for="(type, field) in columnType"
+              :key="field"
+              cols="2"
+            >
+              <v-select
+                v-model="columnType[field]"
+                :label="field"
+                :items="multinetTypes"
+              />
+            </v-col>
+          </v-row>
         </v-card-text>
 
         <v-divider />
@@ -102,6 +116,8 @@ import api from '@/api';
 import { FileType } from '@/types';
 import { validFileType, fileName as getFileName } from '@/utils/files';
 
+type MultinetType = 'label' | 'category' | 'number' | 'date';
+
 const defaultKeyField = '_key';
 export default Vue.extend({
   name: 'TableDialog',
@@ -129,6 +145,7 @@ export default Vue.extend({
       tableCreationError: null as string | null,
       key: defaultKeyField,
       overwrite: false,
+      columnType: {} as { [key: string]: MultinetType },
       uploading: false,
       uploadProgress: null as number | null,
     };
@@ -141,13 +158,24 @@ export default Vue.extend({
         || !!this.fileUploadError
       );
     },
+
+    multinetTypes(): MultinetType[] {
+      return [
+        'label',
+        'category',
+        'number',
+        'date',
+      ];
+    },
   },
+
   methods: {
     restoreKeyField() {
       this.key = defaultKeyField;
     },
 
-    handleFileInput(file: File) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    handleFileInput(this: any, file: File) {
       this.file = file;
 
       if (!file) {
@@ -198,8 +226,7 @@ export default Vue.extend({
           });
         },
 
-        complete() {
-          type MultinetType = 'label' | 'category' | 'number' | 'date';
+        complete: () => {
           const typeRecs = new Map<string, MultinetType>();
 
           columnTypes.forEach((entry, field) => {
@@ -219,7 +246,10 @@ export default Vue.extend({
             typeRecs.set(field, rec);
           });
 
-          console.log(typeRecs);
+          this.columnType = {};
+          typeRecs.forEach((type, field) => {
+            this.$set(this.columnType, field, type);
+          });
         },
       });
     },
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -177,7 +177,7 @@ export default Vue.extend({
           Object.keys(data).forEach((key: string) => {
             if (!columnTypes.has(key)) {
               columnTypes.set(key, {
-                strings: new Set(),
+                strings: new Set<string>(),
                 number: 0,
                 date: 0,
                 total: 0,
@@ -199,7 +199,27 @@ export default Vue.extend({
         },
 
         complete() {
-          console.log(columnTypes);
+          type MultinetType = 'label' | 'category' | 'number' | 'date';
+          const typeRecs = new Map<string, MultinetType>();
+
+          columnTypes.forEach((entry, field) => {
+            const category = entry.strings.size <= 10;
+            const number = entry.number === entry.total;
+            const date = entry.date === entry.total;
+
+            let rec: MultinetType = 'label';
+            if (category && !number && !date) {
+              rec = 'category';
+            } else if (date) {
+              rec = 'date';
+            } else if (number) {
+              rec = 'number';
+            }
+
+            typeRecs.set(field, rec);
+          });
+
+          console.log(typeRecs);
         },
       });
     },
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -77,11 +77,12 @@
             <v-col
               v-for="(type, field) in columnType"
               :key="field"
-              cols="2"
+              cols="3"
             >
               <v-select
                 v-model="columnType[field]"
                 outlined
+                dense
                 :label="field"
                 :items="multinetTypes"
               />
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -81,6 +81,7 @@
             >
               <v-select
                 v-model="columnType[field]"
+                outlined
                 :label="field"
                 :items="multinetTypes"
               />
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -229,8 +229,10 @@ export default Vue.extend({
             entry.strings.add(value);
 
             // See if the value can be converted to a number. Empty strings
-            // convert to 0, so those are excluded specifically.
-            if (value !== '' && !Number.isNaN(Number(value))) {
+            // convert to 0, so those are excluded specifically. Trim the string
+            // here and not above to preserve its literal string value in case
+            // it's not a number.
+            if (value.trim() !== '' && !Number.isNaN(Number(value.trim()))) {
               entry.number += 1;
             }
 
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -112,14 +112,10 @@
 
 <script lang="ts">
 import Vue, { PropType } from 'vue';
-import Papa, { ParseResult } from 'papaparse';
-import dayjs from 'dayjs';
 
 import api from '@/api';
-import { FileType } from '@/types';
-import { validFileType, fileName as getFileName } from '@/utils/files';
-
-type MultinetType = 'label' | 'category' | 'number' | 'date';
+import { FileType, CSVColumnType } from '@/types';
+import { validFileType, fileName as getFileName, csvFileTypeRecommendations } from '@/utils/files';
 
 const defaultKeyField = '_key';
 export default Vue.extend({
@@ -148,7 +144,7 @@ export default Vue.extend({
       tableCreationError: null as string | null,
       key: defaultKeyField,
       overwrite: false,
-      columnType: {} as { [key: string]: MultinetType },
+      columnType: {} as { [key: string]: CSVColumnType },
       uploading: false,
       uploadProgress: null as number | null,
     };
@@ -162,7 +158,7 @@ export default Vue.extend({
       );
     },
 
-    multinetTypes(): MultinetType[] {
+    multinetTypes(): CSVColumnType[] {
       return [
         'label',
         'category',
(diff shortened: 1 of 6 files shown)


src/components/TableDialog.vue
```diff
@@ -140,13 +159,24 @@ export default Vue.extend({
...
     },
+
+    multinetTypes(): MultinetType[] {
+      return [
+        'label',
+        'category',
+        'number',
+        'date',
+      ];
+    },
```
AlmightyYakob, line 170: Why is this in computed? Nothing is changing here so I don't see a reason not to put it in `data`.

src/components/TableDialog.vue
```diff
@@ -157,6 +187,88 @@ export default Vue.extend({
...
+                number: 0,
+                date: 0,
+                total: 0,
+              });
+            }
+
+            const entry = columnTypes.get(key);
+            if (entry === undefined) {
+              throw new Error('impossible');
+            }
```
AlmightyYakob, line 219: If you're only throwing an error here to satisfy TypeScript, and you're okay with betting that this is actually impossible, you can do the following:

```suggestion
            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
            const entry = columnTypes.get(key)!;
```

waxlamp:
@jtomeck, this is the start of a feature that we'll want your help in fixing up the design for later. No need to do anything now, besides maybe take a look at what I've done so you have an idea of where my head is on this feature.

src/components/TableDialog.vue
```diff
@@ -157,6 +187,88 @@ export default Vue.extend({
...
+            const value = data[key];
+            entry.total += 1;
+
+            // Test for categoriness by counting up all the unique strings in
+            // the column.
+            entry.strings.add(value);
+
+            // See if the value can be converted to a number. Empty strings
+            // convert to 0, so those are excluded specifically.
+            if (value !== '' && !Number.isNaN(Number(value))) {
```
AlmightyYakob, line 232: It really doesn't matter but the only way for a string to be falsy is if it's empty, so you could do this
```suggestion
            if (value && !Number.isNaN(Number(value))) {
```

src/components/TableDialog.vue
```diff
@@ -157,6 +187,88 @@ export default Vue.extend({
...
+
+          columnTypes.forEach((entry, field) => {
+            const isKey = field === '_key' || field === '_from' || field === '_to';
+            const category = entry.strings.size <= 10;
+            const number = entry.number === entry.total;
+            const date = entry.date === entry.total;
+
+            let rec: MultinetType = 'label';
+            if (isKey) {
+              rec = 'label';
```
AlmightyYakob, line 254: Why have this initial if statement? Won't this will always be a noop?

AlmightyYakob: Also, what is `rec` supposed to mean? I think either a better variable name or a comment explaining this would help

src/components/TableDialog.vue
```diff
@@ -157,6 +187,88 @@ export default Vue.extend({
...
+              rec = 'date';
+            }
+
+            typeRecs.set(field, rec);
+          });
+
+          this.columnType = {};
+          typeRecs.forEach((type, field) => {
+            this.$set(this.columnType, field, type);
+          });
```
AlmightyYakob, line 269: I think we could move this entire function to a utils file or something, so that it's not cluttering this component so heavily. It could instead return `typeRecs`, at which point you'd perform the highlighted operations in this component.

AlmightyYakob (changes requested):
Overall looks good, and I really like this implementation! Just a few suggestions

src/components/TableDialog.vue
```diff
@@ -157,6 +187,88 @@ export default Vue.extend({
...
+        },
+
+        complete: () => {
+          const typeRecs = new Map<string, MultinetType>();
+
+          columnTypes.forEach((entry, field) => {
+            const isKey = field === '_key' || field === '_from' || field === '_to';
+            const category = entry.strings.size <= 10;
+            const number = entry.number === entry.total;
+            const date = entry.date === entry.total;
```
JackWilb, line 250: What about missing values here? That would mean that a column with all numbers and one missing value would be set to 'label' below.

src/components/TableDialog.vue
JackWilb: I think `rec` is short for recommendation. I think that's reasonable. 

Also, the first statement has to exists, because the 4 boolean expressions above can all evaluate to false, which would cause issues. The default if we can't figure out the type should probably be label

src/components/TableDialog.vue
JackWilb: That's not true (I think Roni mentioned this on slack). The string '0' is falsy

JackWilb:
Also, @waxlamp are there libraries we could leverage for finding the types of the data instead of rolling our own? Or do we have a unique use case and types that we're searching for?

src/components/TableDialog.vue
AlmightyYakob: > I think `rec` is short for recommendation. I think that's reasonable.

I agree.

> Also, the first statement has to exists, because the 4 boolean expressions above can all evaluate to false, which would cause issues. The default if we can't figure out the type should probably be label

This is the section I commented on:

```
let rec: MultinetType = 'label';
if (isKey) {
  rec = 'label';
}
```

I'm saying the `if` statement there is redundant, as `'label'` is already the default value of `rec`.

src/components/TableDialog.vue
AlmightyYakob: I don't think that's true. If you evaluate the expression `!!'0'` in your browser console, the result is `true`. https://developer.mozilla.org/en-US/docs/Glossary/Falsy

JackWilb: I stand corrected. I thought that `'0' == false` (which evaluates to true) implied falsy, but that's just js doing its thing with the double equals. Your change seems appropriate, @AlmightyYakob

src/components/TableDialog.vue
```diff
@@ -72,6 +72,20 @@
...
           </v-row>
+
+          <v-row>
+            <v-col
+              v-for="(type, field) in columnType"
+              :key="field"
+              cols="2"
+            >
+              <v-select
+                v-model="columnType[field]"
```
jtomeck, line 83:
```suggestion
                v-model="columnType[field]"
                outlined
```

jtomeck: Add outlined prop like all the other text fields in the app

jtomeck (commented):
Besides the code changes I suggested, I wanted to ask. If you click the "x" to clear the file you selected, should those inputs disappear again?

src/components/TableDialog.vue
waxlamp: This is a habit of mine when I want to make sure the value being specified cannot be changed by accident. It's sort of the `const` of Vue components.

src/components/TableDialog.vue
waxlamp: Even though mine is more verbose, I think I'll keep it. In this case, using the non-null assertion operator makes sense, but I do think it's somewhat harder to read, and the noise caused by having to suppress a rule seems to also go against legibility.

src/components/TableDialog.vue
waxlamp: Nice catch. I think the exact confusion in this thread is the reason I like to be more explicit with this sort of test. I think it leaves the test more comprehensible by future readers as well.

However, you've made me realize there's a bug in the code: I should really be working with `value.trim()` instead of `value` to avoid having to worry about whitespace-only strings (which convert to `0` but are truthy).

I will add a commit to trim the value we're working with here.

src/components/TableDialog.vue
waxlamp: In that case we'd have no choice but to recommend `label`--we can't coerce all the values to the target type.

src/components/TableDialog.vue
waxlamp: `rec` was meant to be short for "recommendation", but `recommendation` seems too long for a variable name. If it's needed I can use `recommendation` or a different name for this variable.

I kept the `isKey` test for legibility. In pre-typescript days, I wouldn't have assigned any value to `rec` in its initializer, and then the if-elseif-else chain would be explicit and non-redundant. My choices with typescript are to (1) leave `rec` undefined and then type it as `MultinetType | undefined`; (2) cast the if-elseif-else chain as a function that gets called in-place (replacing the `rec =` parts with `return`) and assigned to `rec`; or (3) initialize the variable because I have to, and then use the expclicit if-elseif-else chain as I did.

(1) is a bad choice because it's going out of its way to do something simple.
(2) is a bad choice because it's confusing/ugly.
(3) is non-optimal because it looks redundant. HOWEVER, I think "optimizing" this code would be worse: it would leave it unclear that we deliberately select `label` for Multinet key fields.

This is another case of explicit code for the goal of comprehensibility. Let me know if this is ok or if you really want a different solution here.

src/components/TableDialog.vue
waxlamp: Which entire function do you mean? Looking this over, it's a giant call to `Papa.parse()`, with the different callbacks being sort of related to each other in how this operates. I could see abstracting this whole call into a function outside the component definition but I don't see the point of moving just the type heuristic out, or of moving it to a global place like a utils file.

Let me know a bit more about what you're proposing and I can give you fuller thoughts on it.

waxlamp:
> Also, @waxlamp are there libraries we could leverage for finding the types of the data instead of rolling our own? Or do we have a unique use case and types that we're searching for?

I am pretty sure there are, but you can see how simple the heuristics are that I chose to put into this PR. Keep in mind that the purpose is just to make a recommendation, and let the user make the final decision. We should do some research into type-wrangling libraries in case they are useful in the future, but I'm not sure there's a call for one in this PR.

src/components/TableDialog.vue
JackWilb: We can't convert to a type with some missing values for blank rows? For example, having a number column with 5 missing values. This is standard in most data manipulation libraries/tools

waxlamp: This is a good point. So far we've assumed that data landing in Multinet is "clean" in this sense (i.e., no missing values, etc.). I am not sure if that assumption is code-enforced on the server (@AlmightyYakob probably knows the validation routines best). But whether it is or not, this deserves a deeper discussion. Would you mind filing an issue so we can figure this out the right way?

src/components/TableDialog.vue
waxlamp: Thanks for this; updated.

waxlamp:
> Besides the code changes I suggested, I wanted to ask. If you click the "x" to clear the file you selected, should those inputs disappear again?

They should; currently (as you've pointed out) they do not, and I also get some sort of failure in the console. Thanks for bringing this up; I will fix.

src/components/TableDialog.vue
waxlamp: See 21a337222bfea804dc0734e5b7bba5a6132d7643.

waxlamp:
> > Besides the code changes I suggested, I wanted to ask. If you click the "x" to clear the file you selected, should those inputs disappear again?
> 
> They should; currently (as you've pointed out) they do not, and I also get some sort of failure in the console. Thanks for bringing this up; I will fix.

Fixed in b9a73979370f90080a26d8f6cb26dc17b300ec8b.

I also made the dropdowns a bit wider, and used the `dense` property on them (see f06ef6fed660eb352847798a388aeff04ef6e896).

jtomeck (approved):

src/components/TableDialog.vue
AlmightyYakob: I was missing the fact that the first `if` statement (that I claimed was redundant) has an effect, which is that it sets the priority of the checked conditions (`isKey` is the first checked condition, and if true will ensure that no other conditional statement overwrites `rec`). You could change the redundant assignment with the code below, but I'm also okay with keeping it as is. Up to you.
```
            let rec: MultinetType = 'label';
            if (!isKey) {
              if (category && !number && !date) {
                rec = 'category';
              } else if (number) {
                rec = 'number';
              } else if (date) {
                rec = 'date';
              }
            }
```

src/components/TableDialog.vue
AlmightyYakob: I think it's a bit of an anti-pattern, but I see the benefit of disallowing reassignment. I'm okay with this.

src/components/TableDialog.vue
AlmightyYakob: See #131 for my recommendation.

src/components/TableDialog.vue
AlmightyYakob: I think most of the server-side validation is focused on tables matching each other (e.g. referenced keys existing), not so much column specifics. 

I think Jack's point is valid, one trivial way to address this would be to keep track of empty strings (kept track of in `TypeScore.empty`), and generate a type recommendation if that type count summed with the empty count equals the total. Example with number:

```
const number = entry.number + entry.empty === entry.total;
```

However this may be out of scope if we're just doing a first pass.

src/components/TableDialog.vue
waxlamp: Since this is our test case for transitioning to composition API, it will become moot since in that context we will just use a `const` declaration. So I'll let this stand for now with the understand that it will be "fixed" in that manner.

src/components/TableDialog.vue
waxlamp: I sort of like this approach but in the interest of keeping things moving, let's leave the code as is.

src/components/TableDialog.vue
waxlamp: Approved #131. So I'll wait for you to merge it, and then we can conduct final reviews on this PR.

src/components/TableDialog.vue
waxlamp: I like this idea but yeah it will need to wait until we decide on the semantics/practice of handling missing values in our API. I've linked this discussion into https://github.com/multinet-app/multinet-rfcs/issues/16.

AlmightyYakob (approved):

src/types.ts
```diff
@@ -1,5 +1,7 @@
 import { UploadType } from 'multinet';
 
+export type CSVColumnType = 'label' | 'category' | 'number' | 'date';
```
JackWilb, line 3: Looking at the data types [here](https://www.arangodb.com/docs/stable/aql/fundamentals-data-types.html) to see what types we could natively store in Arango, it looks like an easy one to add would be boolean. However, if we're not really planning to leverage the native types, then this point is irrelevant.

Another quick note, some of these data types in the docs above would be better supported by a JSON upload (for example Arrays and Objects), if we do decide to go for some native storage in Arango.

AlmightyYakob: We have to be careful about this, because I think there's really two types flying around, which happen to intersect here. One is the actual data type that would be used to represent something in ArangoDB (JSON types: null, number, string, bool, etc.). The other is the "semantic" type, that is, how we choose to interpret the data in a column, regardless of it's actual data type. A good example of this might be if a a column is labeled as `date`, but each row entry could represent this with a number (unix timestamp) or a string (ISO 8601). Both would be valid ways to represent this data.

Maybe I'm incorrect about this, but regardless, I think more discussion is needed to flesh out the exact desired behavior. Thoughts @JackWilb @waxlamp?

JackWilb: Right, thanks for explicitly spelling that out for us here. I agree that there are 2 different sets of types that we're talking about.

I guess the types that we use when storing data in Arango are actually irrelevant here since we can make a mapping from any type in the db to the types that we consider to be our "semantic" types.

waxlamp: Good discussion. I like @AlmightyYakob's framing in terms of storage types and semantic types. And I like @JackWilb's observation that "semantic" types are more or less a hint to the vis app writer on what they may wish to do with the raw values coming from the database.

On the specific issue of a "bool" type: it is a good idea, but the difficulty is in execution. For instance, our heuristic could detect a column with only two values in it. Ordinarily, this would be marked as "categorical", but if the two values are strings `"true"` and `"false"` or `"yes"` and `"no"` then we might say it's a "bool" instead. Would we convert these values to literal bools in the database though? My gut says "no". For something like that, my intuition is that we leave it as "categorical" and let the client interpret the values. To have a first-class "bool" type we would really need to think about it (and we'd need a compelling use case first).

Good thoughts though!

waxlamp: Oh, and on the topic of dates: one choice would be to do what BSON does, which is to serialize dates as this weird sort of structure:

```json
{
  "$date": {
    "value": "2020-10-16 16:41:21"
  }
}
```

(that's not precisely correct but it gets the idea across.)

If we sent up values like this from the client, that could serve as a decent signal that this is supposed to be a date.

I don't love this solution but our discussion made me think about BSON. So, just an idea to consider.

waxlamp: > To have a first-class "bool" type we would really need to think about it (and we'd need a compelling use case first).

Just to follow up here, we do now have a compelling use case: UpSet. Let's talk about the boolean semantic type at our meeting on Wednesday.

PR #128, opened 2020-10-13
Add Vue Composition API

AlmightyYakob:
This enables us to use the composition API in the application in general, with a specific example conversion of the `TableDialog` component into composition API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "test:server:up": "cd .. && PIPENV_DONT_LOAD_ENV=1 pipenv run test-server-up"
   },
   "dependencies": {
+    "@vue/composition-api": "^1.0.0-beta.16",
     "axios": "^0.18.1",
     "core-js": "^2.6.5",
     "dayjs": "^1.9.1",
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -120,6 +120,7 @@ import { FileType, CSVColumnType } from '@/types';
 import { validFileType, fileName as getFileName, csvFileTypeRecommendations } from '@/utils/files';
 
 const defaultKeyField = '_key';
+const multinetTypes: readonly CSVColumnType[] = ['label', 'category', 'number', 'date'];
 const fileTypes: readonly FileType[] = [
   {
     extension: ['csv'],
@@ -140,7 +141,6 @@ export default defineComponent({
   setup(props, { emit }) {
     // Type recommendation
     const columnType: Ref<{[key: string]: CSVColumnType}> = ref({});
-    const multinetTypes: readonly CSVColumnType[] = ['label', 'category', 'number', 'date'];
 
     // File selection
     const selectedFile = ref<File | null>(null);
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -54,7 +54,7 @@
           <v-row no-gutters>
             <v-col cols="4">
               <v-text-field
-                v-model="key"
+                v-model="keyField"
                 label="Key Column"
                 append-icon="restore"
                 outlined
@@ -229,7 +229,7 @@ export default defineComponent({
       handleFileInput,
       createTable,
       restoreKeyField,
-      key: keyField,
+      keyField,
       overwrite,
     };
   },
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -111,130 +111,125 @@
 </template>
 
 <script lang="ts">
-import Vue, { PropType } from 'vue';
+import {
+  defineComponent, ref, Ref, computed,
+} from '@vue/composition-api';
 
 import api from '@/api';
 import { FileType, CSVColumnType } from '@/types';
 import { validFileType, fileName as getFileName, csvFileTypeRecommendations } from '@/utils/files';
 
 const defaultKeyField = '_key';
-export default Vue.extend({
-  name: 'TableDialog',
+const fileTypes: FileType[] = [
+  {
+    extension: ['csv'],
+    queryCall: 'csv',
+    hint: 'Comma Separated Value file',
+    displayName: 'CSV',
+  },
+];
 
+export default defineComponent({
+  name: 'TableDialog',
   props: {
     workspace: {
-      type: String as PropType<string>,
+      type: String,
       required: true,
     },
   },
-  data() {
-    return {
-      tableDialog: false,
-      types: [
-        {
-          extension: ['csv'],
-          queryCall: 'csv',
-          hint: 'Comma Separated Value file',
-          displayName: 'CSV',
-        },
-      ] as FileType[],
-      file: null as File | null,
-      fileName: null as string | null,
-      fileUploadError: null as string | null,
-      tableCreationError: null as string | null,
-      key: defaultKeyField,
-      overwrite: false,
-      columnType: {} as { [key: string]: CSVColumnType },
-      uploading: false,
-      uploadProgress: null as number | null,
-    };
-  },
-  computed: {
-    createDisabled(): boolean {
-      return (
-        !this.file
-        || !this.fileName
-        || !!this.fileUploadError
-      );
-    },
+  setup(props, { emit }) {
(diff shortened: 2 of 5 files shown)


src/components/TableDialog.vue
```diff
@@ -95,109 +95,109 @@
...
+    displayName: 'CSV',
+  },
+];
 
+export default defineComponent({
+  name: 'TableDialog',
   props: {
     workspace: {
-      type: String as PropType<string>,
+      type: String,
```
JackWilb, line 118: Is there a difference between `string` and `String`? Why do we prefer the capitalized version here?

JackWilb (commented):
Looks great, I just had one question. I'll leave this to Roni to give a final review, since he has a bit more understanding of the composition API than I do

src/components/TableDialog.vue
waxlamp: `String` is a JavaScript type, while `string` is a TypeScript type.

The `props` declaration requires a JavaScript type to be given here, while the `as` bit allows that JavaScript type to be cast to a pure TypeScript type (using the `PropType` wrapper). This isn't specifically useful here (though, personally, I would probably just keep it), but it is in situations where a JavaScript type like `Object` provides much less information than a TypeScript type like `{ foo: string; bar: number; }`.

AlmightyYakob: `String` is the string class constructor, while `string` is the TypeScript string type. While it may look like we're typing the prop here (I mean semantically that _is_ what we're doing), this is an object, and so whatever is on the right side needs to be a value, specifically a constructor ([relevant](https://vuejs.org/v2/guide/components-props.html#Prop-Types) Vue docs), so that Vue can do it's thing at runtime. It didn't seem like TypeScript needed me to specifically cast this as `PropType<string>` anymore, so I removed that part of it.

AlmightyYakob: Whoops looks like we commented at the same time, but I think we both answered your question.

JackWilb: Thanks for the explanations

waxlamp: @AlmightyYakob's explanation is more accurate than mine, since JavaScript doesn't *really* have types. Mine was more of a gloss over what's really happening.

AlmightyYakob:
@waxlamp @JackWilb PTAL, this has been rebased and is good to go. I really like how this is looking.

AlmightyYakob:
Also, I'd say looking at the diff of `TableDialog` isn't super useful, since it's such a change. It might make sense to just look at the file itself :man_shrugging:

JackWilb (commented):
LGTM. I like that it's much easier to follow what's happening than the other API we've been using. I'll defer to Roni for a final review

package.json
```diff
@@ -19,6 +19,7 @@
     "test:server:up": "cd .. && PIPENV_DONT_LOAD_ENV=1 pipenv run test-server-up"
   },
   "dependencies": {
+    "@vue/composition-api": "^1.0.0-beta.16",
```
waxlamp, line 22: Do you think we're ok to use this prerelease version?

src/components/TableDialog.vue
```diff
@@ -111,130 +111,125 @@
...
-      this.key = defaultKeyField;
-    },
-
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    async handleFileInput(this: any, file: File | undefined) {
-      this.file = file || null;
+  setup(props, { emit }) {
+    // Type recommendation
+    const columnType: Ref<{[key: string]: CSVColumnType}> = ref({});
+    const multinetTypes: readonly CSVColumnType[] = ['label', 'category', 'number', 'date'];
```
waxlamp, line 143: I wonder if this should be hoisted to the top level. (Totally fine to leave it here now, I'm really just wondering.)

src/components/TableDialog.vue
```diff
@@ -111,130 +111,125 @@
...
+
+    // File selection
+    const selectedFile = ref<File | null>(null);
+    const fileName = ref<string | null>(null);
+    const fileUploadError = ref<string | null>(null);
+    async function handleFileInput(file: File) {
```
waxlamp, line 149: Apparently this is allowed to be `undefined` in line 152 (going by the code this line replaced). I think the type of `file` should be `file | undefined`, and the test in 152 should probably be explicit for `undefined`.

(Feel free to leave the test in 152 as it is, since it was like that before. But the type should definitely be changed, and if you changed the test to be more explicit, I think that would actually be an improvement.)

src/components/TableDialog.vue
```diff
@@ -111,130 +111,125 @@
...
+      fileUploadError,
+      tableDialog,
+      tableCreationError,
+      uploading,
+      uploadProgress,
+      createDisabled,
+      handleFileInput,
+      createTable,
+      restoreKeyField,
+      key: keyField,
```
waxlamp, line 230: To try to stick with the standard practice I think this should be changed to `keyField`, and the use of `key` in the template or wherever else it appears should be updated as well.

yarn.lock
```diff
@@ -8609,6 +8616,11 @@ tslib@^1.8.0, tslib@^1.8.1, tslib@^1.9.0:
   resolved "https://registry.yarnpkg.com/tslib/-/tslib-1.13.0.tgz#c881e13cc7015894ed914862d276436fa9a47043"
   integrity sha512-i/6DQjL8Xf3be4K/E6Wgpekn5Qasl1usyw++dAA35Ue5orEn65VIxOA+YvNNl9HV3qv70T7CNwjODHZrLwvd1Q==
 
+tslib@^2.0.1:
```
waxlamp, line 8619: Interesting: https://github.com/microsoft/tslib. Do you happen to know why this gets pulled in by (presumably) `@vue/composition-api`?

waxlamp (approved):
This looks fantastic, a breath of fresh air. Thanks for taking the initiative on this, @AlmightyYakob.

@JackWilb, you already expressed your approval, but does all of the usage here make sense to you? Do you want to study this a bit further and ask any questions on this thread?

src/components/TableDialog.vue
JackWilb: This is the only question I had, too. Is it just reactive things that should be set inside the `setup()` function and everything static should be moved into the top level (outside of the `defineComponent`)?

JackWilb:
> @JackWilb, you already expressed your approval, but does all of the usage here make sense to you? Do you want to study this a bit further and ask any questions on this thread?

Yeah, the usage makes sense. I just had one question about non-reactive things

src/components/TableDialog.vue
waxlamp: There's no rule about non-reactive things not being declared inside `setup()`. In fact, in general, you may want to track some values only during setup, and not expose them to the outside.

In this case, maintaining a list of semantic types that the application knows about might make sense. But since right now this is the only component that cares, we could leave it like this as well.

package.json
AlmightyYakob: I believe there's some weirdness in the pre-beta versions, and they've been on betas for four months now. I'm personally okay with it, but we might want to pin this instead of allowing for minor and patch (and therefore any beta version) bumps.

src/components/TableDialog.vue
AlmightyYakob: You're right, it should be outside of setup. I originally considered placing it outside of setup, but incorrectly thought since I was returning it to the template, that it should be in setup (don't know why I was thinking that tbh). I'll fix it.

src/components/TableDialog.vue
AlmightyYakob: You're right, good catch. I'll fix it.

yarn.lock
AlmightyYakob: Yep, [looks](https://github.com/vuejs/composition-api/blob/0796912c37d2d4389331e6e029b8b270b76458c4/package.json#L74) like it is `@vue/composition-api`.

waxlamp (approved):

PR #130, opened 2020-10-14
Upgrade @vue/cli-service to fix node-forge vuln

JackWilb:
This modifies the selfsigned version to use 1.10.8, which uses the new version of node-forge.

This type of modification of the yarn.lock file by hand is not recommended and I wonder if we should just do the yarn upgrade command that I originally had.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7635,8 +7635,8 @@ select-hose@^2.0.0:
 
 selfsigned@^1.10.7:
   version "1.10.8"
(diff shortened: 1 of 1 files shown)


waxlamp (commented):
This feels like way too big a change set, when the recommended [remediation](https://github.com/multinet-app/multinet-client/network/alert/yarn.lock/node-forge/open) is literally just one line.

I'd prefer to use that remediation directly rather than the yarn upgrade command.

waxlamp (approved):
I was concerned about the `^` in `^0.10.0` for the `node-forge` entries, but [this update in `selfsigned`](https://github.com/jfromaniello/selfsigned/compare/v1.10.7...master) shows that it is actually correct.

Thanks for taking care of this, @JackWilb.

PR #132, opened 2020-10-20
Update use of `appregistry.json`

AlmightyYakob:
I encountered this a bit when trying to run these applications locally, and thought I could improve the user experience a bit. Now, the app registry can be changed locally as needed, without needing to change checked-in files.


JackWilb (commented):
Overall this looks great. I wonder if we should specify a default dev file that has the ports for the vis apps locked (I'm thinking we should use 8081 for the matrix and 8082 for the nodelink). That would make it even quicker to spin up a dev environment. Although, that said, to do this step once is not a big deal

.gitignore
```diff
@@ -24,3 +24,5 @@ yarn-error.log*
 # Deployment files
 git-sha
 ga-tag
+
+appregistry.json
```
waxlamp, line 28: I think this will ignore the file everywhere in the repo, whereas really we just want to ignore the file in its new location at `src/assets/appregistry.json`, right?

netlify.toml
```diff
@@ -1,6 +1,6 @@
 [build]
 publish = "dist"
-command = "yarn build"
+command = "yarn build:prod"
```
waxlamp, line 3: Is there any reason for separate commands here? In Vue applications, `serve` is for dev and `build` is for prod already. If the danger is that you don't want to overwrite the app registry, I think we can update the `cp` command in the `init-appregistry` script.

package.json
```diff
@@ -5,6 +5,8 @@
   "scripts": {
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
+    "build:prod": "yarn init-appregistry && yarn build",
+    "init-appregistry": "cp src/assets/prod.appregistry.json src/assets/appregistry.json",
```
waxlamp, line 9: Two comments about this:
1. You can use the `-n` ("no clobber") option to prevent overwriting a file that has already been copied here. Then we can include an `init-appregistry:force` that looks like this line (i.e., *does* overwrite the destination file). I just tested it, and when the file already exists, `cp` still exits with success. That would allow the `yarn build` script to work safely without having to split it into a `build` and a `build:prod` script.
2. It's a bit risky to use a Linux command line program here, since this won't work on e.g. Windows. Now, we don't really support Windows, so I'm ok with this at least for now, but there are :nauseated_face: NPM-based cross-platform solutions instead. I wouldn't look into any of those for now, but let's open an issue to track this possible future pitfall.

waxlamp (changes requested):
This is really nice, thanks for taking the initiative, @AlmightyYakob.

I have a couple of suggestions, let me know what you think before we proceed.

waxlamp:
> Overall this looks great. I wonder if we should specify a default dev file that has the ports for the vis apps locked (I'm thinking we should use 8081 for the matrix and 8082 for the nodelink). That would make it even quicker to spin up a dev environment. Although, that said, to do this step once is not a big deal

This is a good idea in general, but let's think about it a bit more. I don't like the idea of locking it to specifically 8081 and 8082 (even though those are the common case that will pop up, for anyone who is already running a Vue application in dev with the same default parameters we use, the port numbers will be different). I wonder if we could just write a small utility that would update the URLs with local port numbers (or revert them to the original values). That wouldn't be too hard. I may take a crack at it as a way to take a break from my proposal deadline.

.gitignore
AlmightyYakob: It does ignore that file everywhere, but I'd wager that's what we want. As in, I don't think we have a reason to have a separate `appregistry.json` somewhere else, and ignoring it everywhere means the `.gitignore` won't change if we decide to change the location again.

AlmightyYakob:
> This is a good idea in general, but let's think about it a bit more. I don't like the idea of locking it to specifically 8081 and 8082 (even though those are the common case that will pop up, for anyone who is already running a Vue application in dev with the same default parameters we use, the port numbers will be different). I wonder if we could just write a small utility that would update the URLs with local port numbers (or revert them to the original values). That wouldn't be too hard. I may take a crack at it as a way to take a break from my proposal deadline.

I had a similar idea, but thought it would be out of scope for this PR. Just to clarify, you mean somehow integrating the applications such that the appregistry of the client is updated whenever one of the apps is running locally? If that's the case, we could probably create a bash script in the client (and expose it in the `package.json`) that runs all three, and does some smart injection/manipulation of the appregistry.

I'm not really sure what the best solution is, but we'd have to be careful not to dig ourselves a hole if we decide to approach this. Like you said, I'll leave this task to you.

package.json
AlmightyYakob: 1. I would say if we're going to use the `-n` flag (which we probably should) in order to just have one `build` script, then there's probably no need for an additional `init-appregistry:force` script. In my mind, if an `appregistry.json` is already there, then that satisfies the "initialize" condition (the condition being, an `appregistry.json` is required for the application to run).

2. Good catch, although I agree we don't really deal with windows at the current moment. I'd say in this case we could probably just punt on this until it actually becomes an issue, since even if it does eventually become an issue, a fix / one-time fix is low cost.

netlify.toml
AlmightyYakob: There's wasn't a super good reason, it just seemed "bad" somehow to change the `build` script, but looking at it now I think changing it would be fine. I'll defer this discussion to the one below.

.gitignore
waxlamp: I think it would be better to be more precise. Then, if you happen to have appregistry.json somewhere where the app is not looking for it, then you will be less confused why your changes aren't appearing, etc. The cost for that is one extra change (to .gitignore) if/when we decide to change where the app does look for this file.

package.json
waxlamp: Agree with both points. So, use `-n` and dump the force script, and don't worry about filing an issue for point (2) (since it will be obvious if and when it pops up).

waxlamp:
> > This is a good idea in general, but let's think about it a bit more. I don't like the idea of locking it to specifically 8081 and 8082 (even though those are the common case that will pop up, for anyone who is already running a Vue application in dev with the same default parameters we use, the port numbers will be different). I wonder if we could just write a small utility that would update the URLs with local port numbers (or revert them to the original values). That wouldn't be too hard. I may take a crack at it as a way to take a break from my proposal deadline.
> 
> I had a similar idea, but thought it would be out of scope for this PR. Just to clarify, you mean somehow integrating the applications such that the appregistry of the client is updated whenever one of the apps is running locally? If that's the case, we could probably create a bash script in the client (and expose it in the `package.json`) that runs all three, and does some smart injection/manipulation of the appregistry.
> 
> I'm not really sure what the best solution is, but we'd have to be careful not to dig ourselves a hole if we decide to approach this. Like you said, I'll leave this task to you.

My idea was just to write a Python script that edits appregistry.json in-place (you would like, supply a sequence of port numbers or something). It's probably not worth doing that when you could also just... edit the file yourself. You'd have to do that at most once per Multinet work session anyway.

The automated bash script idea is intriguing but I think maybe we just let this go for now. We'd be putting engineering effort toward a problem that already has a solution that's not much worse than what-all we're proposing.

.github/workflows/main.yml
```diff
@@ -34,6 +34,9 @@ jobs:
         if: steps.client-cache.outputs.cache-hit != 'true'
         run: yarn install
 
+      - name: Populate app registry
+        run: yarn init-appregistry
+
```
waxlamp, line 39: Probably don't need this anymore, right?

netlify.toml
waxlamp: Should this now remain as `yarn build`?

src/assets/.gitignore
```diff
@@ -0,0 +1 @@
+appregistry.json
```
waxlamp, line 1: Ah I see. I was recommending just listing `src/assets/appregistry.json` in the top-level .gitignore file, but this is fine too.

waxlamp (commented):
Looks good, but just had a few nitpicks before final approval.

.github/workflows/main.yml
AlmightyYakob: We actually need that for a separate reason. This is because if the `appregistry.json` file isn't present, the linting will fail, because it can't import the file. I opted for this instead of switching `yarn build` to run before linting, because if we did that, then we wouldn't only see the build step failures, and not the informative ones from linting.

waxlamp (approved):
Thanks for this improvement, @AlmightyYakob!

JackWilb (approved):

PR #135, opened 2020-11-02
Add a "boolean" semantic type detection

waxlamp:
This PR adds a `boolean` semantic type to the ones detected at CSV upload time.

This works by detecting whether a column has exactly two unique values, and whether those two values are one of several possible pairs of boolean-indicative values. These include the strings `0` and `1` but also several of the YAML-compliant boolean values.

Some things I need feedback on:
- Currently, the heuristic does not allow for mix-and-match between the possible signal words--e.g., a column with `yes` for true and `0` for false, or a mix of `yes` and `true` for true, will NOT be detected as a boolean column. **Should we allow for a mix-and-match approach, or restrict to just a single pair of words?** (If we support "mix-and-match" then the heuristic will need to change slightly, but this is not a big deal.)
- If we do allow mix-and-match, should we support the full range of values given by the [YAML spec](https://yaml.org/type/bool.html)?
- How do we store these values in the database? My gut says we simply translate each word in the column into an Python/Arango-level `bool` value and store that. Then, clients will receive JSON boolean values and can interpret those how they wish.

### DIFF
--- a/src/utils/files.ts
+++ b/src/utils/files.ts
@@ -21,8 +21,16 @@ function isBoolean(strings: Set<string>) {
   const candidates: Array<[string, string]> = [
     ['0', '1'],
     ['no', 'yes'],
+    ['No', 'Yes'],
+    ['NO', 'YES'],
+    ['n', 'y'],
+    ['N', 'Y'],
     ['false', 'true'],
+    ['False', 'True'],
+    ['FALSE', 'TRUE'],
     ['off', 'on'],
+    ['Off', 'On'],
+    ['OFF', 'ON'],
   ];
 
   // Consider the string set to be compatible with the "boolean" semantic type
--- a/src/utils/files.ts
+++ b/src/utils/files.ts
@@ -11,6 +11,26 @@ function validFileType(file: File, allowedTypes: readonly FileType[]) {
   return allowedTypes.some((type) => type.extension.includes(extension) && validUploadType(type.queryCall));
 }
 
+function isBoolean(strings: Set<string>) {
+  // This function tests wehther both elements of an array belong to the string
+  // set.
+  const hasPair = ([a, b]: [string, string]) => strings.has(a) && strings.has(b);
+
+  // This is a list of two-element arrays containing valid false/true value
+  // pairs.
+  const candidates: Array<[string, string]> = [
+    ['0', '1'],
+    ['no', 'yes'],
+    ['false', 'true'],
+    ['off', 'on'],
+  ];
+
+  // Consider the string set to be compatible with the "boolean" semantic type
+  // if it contains exactly two unique values, and that pair of values is
+  // precisely one of the candidate pairs above.
+  return strings.size === 2 && candidates.some(hasPair);
+}
+
 interface TypeScore {
   strings: Set<string>;
   number: number;
@@ -72,6 +92,7 @@ async function csvFileTypeRecommendations(file: File): Promise<Map<string, CSVCo
       complete: () => {
         columnTypes.forEach((entry, field) => {
           const isKey = field === '_key' || field === '_from' || field === '_to';
+          const boolean = isBoolean(entry.strings);
           const category = entry.strings.size <= 10;
           const number = entry.number === entry.total;
           const date = entry.date === entry.total;
@@ -79,6 +100,8 @@ async function csvFileTypeRecommendations(file: File): Promise<Map<string, CSVCo
           let rec: CSVColumnType = 'label';
           if (isKey) {
             rec = 'label';
+          } else if (boolean) {
+            rec = 'boolean';
           } else if (category && !number && !date) {
             rec = 'category';
           } else if (number) {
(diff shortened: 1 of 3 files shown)


src/utils/files.ts
```diff
@@ -11,6 +11,26 @@ function validFileType(file: File, allowedTypes: readonly FileType[]) {
   return allowedTypes.some((type) => type.extension.includes(extension) && validUploadType(type.queryCall));
 }
 
+function isBoolean(strings: Set<string>) {
+  // This function tests wehther both elements of an array belong to the string
```
JackWilb, line 15:
```suggestion
  // This function tests whether both elements of an array belong to the string
```

JackWilb (approved):
Caught a typo.

As for your questions:
> * Currently, the heuristic does not allow for mix-and-match between the possible signal words--e.g., a column with yes for true and 0 for false, or a mix of yes and true for true, will NOT be detected as a boolean column. Should we allow for a mix-and-match approach, or restrict to just a single pair of words? (If we support "mix-and-match" then the heuristic will need to change slightly, but this is not a big deal.)

I think this is an edge case. It's likely that the format will be one of these pairs. I say that for ease of programming (and for not programming for every possible edge case), we should leave it as is. Let's just make sure that in the docs we specify that they must be these pairs that are specified in the code. Worst case is you have 2 distinct values that don't match, they're not considered to be bool (likely they would be category instead).

> * If we do allow mix-and-match, should we support the full range of values given by the YAML spec?

If we did then yes, but I don't think we should.

> * How do we store these values in the database? My gut says we simply translate each word in the column into an Python/Arango-level bool value and store that. Then, clients will receive JSON boolean values and can interpret those how they wish.

If a user specifies that the column is a bool (note that they are the ones who give the "okay" on the type defs in the client) then we should store that in Arango as JSON boolean values. If a user doesn't want that behavior, they should instead select category. Ultimately, this is just to make a suggestion to the user, they have to be the ones to make the final call.

src/utils/files.ts
waxlamp: Thanks!

waxlamp:
> Caught a typo.
> 
> As for your questions:
> 
> > * Currently, the heuristic does not allow for mix-and-match between the possible signal words--e.g., a column with yes for true and 0 for false, or a mix of yes and true for true, will NOT be detected as a boolean column. Should we allow for a mix-and-match approach, or restrict to just a single pair of words? (If we support "mix-and-match" then the heuristic will need to change slightly, but this is not a big deal.)
> 
> I think this is an edge case. It's likely that the format will be one of these pairs. I say that for ease of programming (and for not programming for every possible edge case), we should leave it as is. Let's just make sure that in the docs we specify that they must be these pairs that are specified in the code. Worst case is you have 2 distinct values that don't match, they're not considered to be bool (likely they would be category instead).
> 

Actually I don't think the *programming* would be too much more difficult. But documentation and testing would probably be, so I think for now at least, you are right and we should just detect the case of two valid signal words in order to determine that a column is boolean. And then if we ever decide we want that mix-and-match approach, it shouldn't be difficult to add it.

> > * If we do allow mix-and-match, should we support the full range of values given by the YAML spec?
> 
> If we did then yes, but I don't think we should.

Ok. So, it looks like the PR already has all the boolean word pairs from YAML in it, except `y`/`n`, not accounting for the capitalization variations. I am going to add `y`/`n` for completeness, and I think I might go ahead and add the capitalization variants as well (there are not too many of them).

My main worry here was, if we don't have mix-and-match, whether mixing capitalization should be allowed. But the answer to that, at least for now, should be "no". 

> 
> > * How do we store these values in the database? My gut says we simply translate each word in the column into an Python/Arango-level bool value and store that. Then, clients will receive JSON boolean values and can interpret those how they wish.
> 
> If a user specifies that the column is a bool (note that they are the ones who give the "okay" on the type defs in the client) then we should store that in Arango as JSON boolean values. If a user doesn't want that behavior, they should instead select category. Ultimately, this is just to make a suggestion to the user, they have to be the ones to make the final call.

:+1:

JackWilb (approved):

src/utils/files.ts
```diff
@@ -11,6 +11,34 @@ function validFileType(file: File, allowedTypes: readonly FileType[]) {
...
+    ['No', 'Yes'],
+    ['NO', 'YES'],
+    ['n', 'y'],
+    ['N', 'Y'],
+    ['false', 'true'],
+    ['False', 'True'],
+    ['FALSE', 'TRUE'],
+    ['off', 'on'],
+    ['Off', 'On'],
+    ['OFF', 'ON'],
```
AlmightyYakob, line 33: Is On/Off ever actually used to represent boolean values? I've never seen it used this way.

waxlamp: It's part of the [YAML spec](https://yaml.org/type/bool.html) :shrug: 

If we have the other YAML values here we should have On/Off too.

AlmightyYakob (approved):

PR #136, opened 2020-11-13
Add an upload preview table with column type selectors

### LINKED ISSUE #134: Implement preview table during CSV upload
The intent is to have a mini table view (similar, but not identical to, the existing table view component) that appears in the upload dialog after selecting a file, at the point that the column type dropdowns currently appear. An example of this feature in @kirangadhave's UpSet work can be used for inspiration (but, of course, keeping the style and feel of the component to what is appropriate for Multinet):

Table View

Specific features for our iteration of this:
- condensed so that it feels right inside a modal dialog
- includes the existing `<v-select>`s for type selection into the header of the table

@jtomeck, could you start by creating a mockup of this in Figma (or wherever is most comfortable for you)?

UPDATE: Jared has done a mockup which we can use as a guide for implementation:
preview_table

waxlamp:
This uses @jtomeck's design for a preview table within a stepper to (1) select a CSV file to upload and (2) examine a sample of the data and set up column data types before (3) uploading the table to Multinet.

@jtomeck: I need help making my use of Vuetify have the right styles to match up to your design. For example:
- the colors are wrong
- the `<v-select>`s look ugly as sin
- no finish flag in the stepper
- the stepper got reduced to only two steps

Let me know what you think.

Closes #134

### DIFF
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -47,14 +47,92 @@
 
       <v-stepper-items>
         <v-stepper-content step="1">
-          <h1>hi</h1>
+          <v-card>
+            <v-card>
+              <v-card-text class="px-4 pt-4 pb-1">
+                <v-layout wrap>
+                  <v-flex>
+                    <v-file-input
+                      :error-messages="fileUploadError"
+                      label="Upload File"
+                      prepend-inner-icon="attach_file"
+                      prepend-icon=""
+                      single-line
+                      clearable
+                      dense
+                      outlined
+                      @change="handleFileInput"
+                    />
+                  </v-flex>
+                </v-layout>
+                <v-layout wrap>
+                  <v-flex>
+                    <v-text-field
+                      v-model="fileName"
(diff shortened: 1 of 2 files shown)


jtomeck (changes requested):
I just made a commit improving some of the look and styles. For some reason the table scrolling is not working properly. That went away when I removed the padding of the stepper content around the table. Do any of you see anything that could be wrong? If not I can remove the padding and it should resolve itself, I just like the look better without the padding. @waxlamp @AlmightyYakob

waxlamp:
> I just made a commit improving some of the look and styles. For some reason the table scrolling is not working properly. That went away when I removed the padding of the stepper content around the table. Do any of you see anything that could be wrong? If not I can remove the padding and it should resolve itself, I just like the look better without the padding. @waxlamp @AlmightyYakob

I think this looks great, @jtomeck.

I know there's a problem with scrolling--can we file an issue and fix that as it causes problems later?

jtomeck (approved):

PR #137, opened 2020-11-16
Replace `<v-content>` with `<v-main>`

waxlamp:
This is in response to console warnings explaining that `<v-content>` is deprecated.

@jtomeck, could you verify that this change is the correct way to do things? I couldn't find a migration guide addressing this specifically, so I just followed the advice in the warning message itself (which just says to use `<v-main>` instead).

### DIFF
--- a/src/views/AQLWizard.vue
+++ b/src/views/AQLWizard.vue
@@ -1,6 +1,6 @@
 <template>
   <v-container fluid>
-    <v-content>
+    <v-main>
       <v-navigation-drawer
         right
         app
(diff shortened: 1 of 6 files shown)


jtomeck (approved):
I guess I never really knew what `<v-main>` was for. It always seemed like an additional element, but if it's supposed to be used INSTEAD of content, I think that makes more sense (I used to see `<v-main>` inside of `<v-content>`).

waxlamp:
This discusses v-app and v-main, but I don't really understand it fully: https://vuetifyjs.com/en/features/layouts/.

With this PR though, I no longer get warnings in the console, and the app seems to behave the same as before. So I'll merge this and if we run into problems later we will fix. Thanks @jtomeck!

PR #138, opened 2020-11-19
Add column type information to table creation call

waxlamp:
This PR makes use of https://github.com/multinet-app/multinetjs/pull/22.

This uses the updated multinetjs library to send column type information at table upload time.

Depends on https://github.com/multinet-app/multinetjs/pull/22 being merged and released.
Depends on #136.

### DIFF
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -264,6 +264,7 @@ export default defineComponent({
           data: selectedFile.value,
           key: keyField.value,
           overwrite: overwrite.value,
+          columnTypes: columnType.value,
         }, {
           onUploadProgress: handleUploadProgress,
         });
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -163,7 +163,7 @@ import {
 
 import api from '@/api';
 import { FileType, CSVColumnType } from '@/types';
-import { validFileType, fileName as getFileName, csvFileTypeRecommendations } from '@/utils/files';
+import { validFileType, fileName as getFileName, analyzeCSV } from '@/utils/files';
 
 const defaultKeyField = '_key';
 const multinetTypes: readonly CSVColumnType[] = ['label', 'boolean', 'category', 'number', 'date'];
@@ -221,12 +221,12 @@ export default defineComponent({
         fileUploadError.value = null;
       }
 
-      const typeRecs = await csvFileTypeRecommendations(file);
-      columnType.value = Array.from(typeRecs.typeRecs.keys()).reduce(
-        (acc, key) => ({ ...acc, [key]: typeRecs.typeRecs.get(key) }), {},
+      const analysis = await analyzeCSV(file);
+      columnType.value = Array.from(analysis.typeRecs.keys()).reduce(
+        (acc, key) => ({ ...acc, [key]: analysis.typeRecs.get(key) }), {},
       );
 
-      sampleRows.value = [...typeRecs.sampleRows];
+      sampleRows.value = [...analysis.sampleRows];
     }
 
     // Upload options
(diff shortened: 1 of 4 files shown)


src/utils/files.ts
```diff
@@ -46,10 +46,16 @@ interface TypeScore {
...
 }
 
-async function csvFileTypeRecommendations(file: File): Promise<Map<string, CSVColumnType>> {
-  const parsePromise: Promise<Map<string, CSVColumnType>> = new Promise((resolve) => {
+interface Recommendation {
+  typeRecs: Map<string, CSVColumnType>;
+  rowSample: Array<{}>;
+}
+
+async function csvFileTypeRecommendations(file: File): Promise<Recommendation> {
```
AlmightyYakob, line 54: If we're going to be returning multiple things from here, I feel like we should rename these things. The interface could be something like `CSVAnalysisResult`, and the function something like `AnalyzeCSV`. I'm not sure if those are the best names, but we're doing two different things here now, and I think it'd be good to reflect that in the function/type names.

src/components/TableDialog.vue
```diff
@@ -139,6 +185,22 @@ export default defineComponent({
     },
   },
   setup(props, { emit }) {
+    // Stepper control
+    const step: Ref<number> = ref(1);
+    const rowSample: Ref<Array<{}>> = ref([]);
```
AlmightyYakob, line 190: It's a minor thing, but I think a better name for this is `sampleRows`. I feel like that better conveys that this is just meant to be a sample of rows from the whole file. A comment may be good too.

Feel free to push back if you think it's too minor of a change.

src/components/TableDialog.vue
```diff
@@ -139,6 +185,22 @@ export default defineComponent({
...
+        return [];
+      }
+
+      const keys = Object.keys(rowSample.value[0]);
+
+      return keys.map((key) => ({
+        text: key,
+        value: key,
+      }));
+    });
```
AlmightyYakob, line 202: I think this looks just a bit cleaner:
```suggestion
    const headers = computed(() => {
      const keys = Object.keys(rowSample.value[0] || {});
      return keys.map((key) => ({
        text: key,
        value: key,
      }));
    });
```

AlmightyYakob:
I think this looks really great! Just left a few suggestions

waxlamp:
> Another suggestion I just thought of, which we'll likely want to do as a follow up:
> 
> Since we've now got this type selection UI, we could add a `primaryKey` type to our system, which would replace the following section in the first step:
> 
> Screenshot from 2020-11-19 14-39-59
> 
> We could remove this option from the first step, and just have it as another type in the column drop down in the second step. We would also likely default that to the `_key` field if it's there, or just the first column if it's not there.
> 
> There's important distinction from the rest of the types however, which is that there can only be one primary key column. We'd have to add some logic when it comes to selecting that type, and we'd need to create a new validation error on the server to handle multiple supplied primary keys, but those tasks seem fairly low-effort.
> 
> Since this involves both server and client changes, we should probably address it in a separate issue, and move forward with this PR. However, I thought it was worthy of discussion.
> 
> @waxlamp lmk what you think.

I think this is a good idea and I agree that we should address it in a followup. Can you an issue and link it back here?

src/components/TableDialog.vue
waxlamp: See 137f1671e635636d08d00e3b0da614191b96cc1d.

src/utils/files.ts
waxlamp: Good idea; check out the names I chose: d7ef1fbd31555f78e04484abcb544b653aa5ea4f.

AlmightyYakob (approved):

AlmightyYakob (commented):
While I approve of these changes, I'm holding off my official approval until the proper version of `multinetjs` is listed.

AlmightyYakob (approved):
:shipit:

Issue #57, opened 2020-03-26
Sorting the rows of a table by clicking the variable name in the header is broken


JackWilb:
In a table (both node and edge table) sorting the rows by clicking the variables names at the top doesn't actually sort the table. The sort icon remains but the row order does not change.

PR #145, opened 2020-12-04
Display semantic types in table view

waxlamp:
This PR makes use of https://github.com/multinet-app/multinetjs/pull/23 to display the semantic types in the header row of the table view, if they exist.

@AlmightyYakob, could you make sure the code looks ok?

@jtomeck, could you take a look at design issues? The way I've done it works but I can tell it doesn't look/feel great.

Eventually, after migrations are implemented, we can dump the "if it exists" logic, since we can be confident that the type metadata will always exist.

TODO
- [ ] file issue to get rid of "if it exists" logic
- [x] force push the multinetjs version number after merging and publishing https://github.com/multinet-app/multinetjs/pull/23

### DIFF
--- a/src/views/TableDetail.vue
+++ b/src/views/TableDetail.vue
@@ -185,15 +185,22 @@ export default Vue.extend({
   watch: {
     workspace() {
       this.update();
+      this.updateTypes();
     },
     table() {
       this.update();
+      this.updateTypes();
     },
 
     pagination() {
       this.update();
     },
   },
+
+  mounted() {
+    this.updateTypes();
+  },
+
   methods: {
     rowClassName(index: number): 'even-row' | 'odd-row' {
       return index % 2 === 0 ? 'even-row' : 'odd-row';
@@ -207,8 +214,6 @@ export default Vue.extend({
         offset: (pagination.page - 1) * pagination.itemsPerPage,
         limit: pagination.itemsPerPage,
       });
-
-      this.columnTypes = await api.tableColumnTypes(this.workspace, this.table);
 
       const {
         rows,
@@ -244,6 +249,10 @@ export default Vue.extend({
       this.tables = await api.tables(this.workspace, {
         type: 'all',
       });
+    },
+
+    async updateTypes() {
+      this.columnTypes = await api.tableColumnTypes(this.workspace, this.table);
     },
   },
 });
--- a/src/views/TableDetail.vue
+++ b/src/views/TableDetail.vue
@@ -103,7 +103,9 @@
                   class="pt-2 pb-4"
                 >
                   {{ header.text }}
-                  <span>{{ columnTypes[header.text] ? `(${columnTypes[header.text]})` : "" }}</span>
+                  <span v-if="columnTypes[header.text]">
+                    ({{ columnTypes[header.text] }})
+                  </span>
                 </th>
               </tr>
             </thead>
--- a/src/views/TableDetail.vue
+++ b/src/views/TableDetail.vue
@@ -85,6 +85,7 @@
           height="calc(100vh - 123px)"
           class="table-details"
           :headers="dataTableHeaders"
+          hide-default-header
           :items="dataTableRows"
           :footer-props="{
             itemsPerPageOptions: [10, 20, 50, 100],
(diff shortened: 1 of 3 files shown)


JackWilb:
I just did a quick test on the deploy preview and I wasn't able to see the data type a dataset that I uploaded. Is there a problem somewhere in the pipeline?

src/views/TableDetail.vue
```diff
@@ -225,6 +250,10 @@ export default Vue.extend({
         type: 'all',
       });
     },
+
+    async updateTypes() {
+      this.columnTypes = await api.tableColumnTypes(this.workspace, this.table);
+    },
```
AlmightyYakob, line 256: If this function is called repeatedly just to update the types, shouldn't this just be as computed function? I'm guessing this is due to the call itself being async, but this could be another spot to use `asyncComputed`. Example use below

https://github.com/multinet-app/multinet-client/blob/1cb0b0e35f3a092d887b34afe2f74804c4a211b0/src/components/PermissionsDialog.vue#L295-L306

AlmightyYakob (changes requested):
Looks good, just a couple requests

src/views/TableDetail.vue
waxlamp: I didn't realize we were already using asyncComputed. Have we run into any problems with it?

I'll reformulate this with asycComputed in the meantime.

waxlamp: (or rather, instead of "didn't realize", I should have said "forgot" :face_with_head_bandage:)

waxlamp: See c2eeec158ffab077cc8ee20171effb90a836c4f5.

src/views/TableDetail.vue
```diff
@@ -85,14 +85,32 @@
...
           :options.sync="pagination"
-        />
+        >
+          <template v-slot:header>
+            <thead dark>
+              <tr>
+                <th
+                  v-for="(header, i) in dataTableHeaders"
+                  :key="i"
+                  class="pt-2 pb-4"
```
JackWilb, line 103: It looks a little strange to me that the headers are not centered on the page. I'll defer to Jared though, for final say on this.

```suggestion
                  class="pt-2 pb-2"
```

src/views/TableDetail.vue
```diff
@@ -164,23 +183,45 @@ export default Vue.extend({
...
+      },
+
+      default: {},
+    },
+  },
+
   watch: {
-    workspace() {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    workspace(this: any) {
```
JackWilb, line 203: Why is this now required in way more places?

JackWilb (changes requested):
A couple of design problems, we lost the sort icons somehow and the column labels now seem to be too high (too much bottom padding?).

This is not related to this change set, but shouldn't the internal columns be restricted to label (for example _id, _from, _to, etc.). When I tried to generate a new table on the deploy preview, the column, _id shows no type. This could be an issue for the multinet-server repo.

waxlamp:
> This is not related to this change set, but shouldn't the internal columns be restricted to label (for example _id, _from, _to, etc.). When I tried to generate a new table on the deploy preview, the column, _id shows no type. This could be an issue for the multinet-server repo.

This is a good point. I believe this happens because `_id` is a column that Arango adds itself, so we've sort of forgotten to deal with it. I've opened #146 to track this.

waxlamp:
> A couple of design problems, we lost the sort icons somehow

I didn't mean to do this but it's probably a good change, since the sort icons weren't doing anything. There's an issue to fix them (#57), and along with that we can bring the icons back. This happened because I had to use custom content to include the type annotations with the header name.

> and the column labels now seem to be too high (too much bottom padding?).
> 

I see that now. Do you know how to fix it? This is something @jtomeck should help us with.

src/views/TableDetail.vue
waxlamp: @jtomeck, can you weigh in on this? I was copying this from a [different place in the code with custom headers](https://github.com/multinet-app/multinet-client/blob/master/src/components/TableDialog.vue#L133), but @JackWilb is right, this throws off vertical alignment. With Jack's suggestion, it looks better.

jtomeck (approved):
@waxlamp - I was just working on a commit where I vertical centered the table header text. You beat me to it. I was also thinking that the table header text looked too small, but upon viewing the default size, I think any bigger looks kinda weird, so I think we're fine. Other than missing the sorting toggles, I think this looks good.

src/views/TableDetail.vue
JackWilb: I can see you closed this in 865ef9f and Jared approved. I'll mark as resolved

src/views/TableDetail.vue
JackWilb: Haha, I'm definitely susceptible to nerd sniping, too. Your resolution seems appropriate for now.

JackWilb (dismissed):

AlmightyYakob (changes requested):
While I approve of these changes, I'm holding off on officially approving until https://github.com/multinet-app/multinetjs/pull/23 is merged, and the version here is updated.

AlmightyYakob (approved):

PR #153, opened 2021-03-04
Use OAuth login

zachmullen:
This won't work until https://github.com/multinet-app/multinetjs/pull/28 is merged and a new release of multinetjs is cut.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -3,6 +3,8 @@ import AsyncComputed from 'vue-async-computed';
 import VueCompositionApi from '@vue/composition-api';
 import Vue from 'vue';
 import App from './App.vue';
+import api from './api';
+import oauthClient from './oauth';
 import vuetify from './vuetify';
 import router from './router';
 import store from './store';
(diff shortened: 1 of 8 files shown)


src/main.ts
```diff
@@ -12,10 +14,14 @@ Vue.config.productionTip = false;
...
 
-new Vue({
-  render: (h) => h(App),
-  router,
-  // Necessary to correctly use direct-vuex
-  store: store.original,
-  vuetify,
-}).$mount('#app');
+oauthClient.maybeRestoreLogin().then(() => {
+  Object.assign(api.client.axios.defaults.headers.common, oauthClient.authHeaders);
```
waxlamp, line 18: @zachmullen, how did you know to access the axios instance's persistent request headers in this way? Is this "safe"?

zachmullen: I believe I looked it up and this was the recommended approach. It seems to work -- I'm not sure under exactly what conditions it's safe or not.

AlmightyYakob:
This PR is now dependent on a new release of `multinetjs` being cut. One thing to note is that while if a new release was cut, login would technically work, but not much else would, because the methods in `multinetjs` still need to be updated to access the new API. We'll need to do some coordinated development between `multinetjs`, the client and the API, in order to get things fully working. I'm thinking that while this is happening, we can instead cut pre-releases for any `girder4` stuff, making full releases once it makes sense to. We'd also need to follow a similar process with the viz apps. Thoughts @waxlamp?

waxlamp:
> This PR is now dependent on a new release of `multinetjs` being cut. One thing to note is that while if a new release was cut, login would technically work, but not much else would, because the methods in `multinetjs` still need to be updated to access the new API. We'll need to do some coordinated development between `multinetjs`, the client and the API, in order to get things fully working. I'm thinking that while this is happening, we can instead cut pre-releases for any `girder4` stuff, making full releases once it makes sense to. We'd also need to follow a similar process with the viz apps. Thoughts @waxlamp?

This sounds like exactly the right approach. I will work on cutting a pre-release version of multinetjs today so that this PR works properly, and then we'll spin up Emi on how these things work so that she can begin to do some basic transformations of the client code and multinetjs to shift functionality over to Girder 4.

PR #154, opened 2021-03-04
Add skeleton loaders across the app

### LINKED ISSUE #147: Use skeleton loaders when hydrating with data for the first time
https://vuetifyjs.com/en/components/skeleton-loaders/

For components like the table and network panels, this is preferable to current behavior, which shows the empty state, then flips over to the full state after the data requests resolve.

JackWilb:
Closes #147

Uses skeleton loaders across the app to give a more seamless experience. This new functionality means that the app doesn't load into an empty state and flip to the correct values.

There are a couple of issues I have regarding how many skeleton loaders to use for each piece, and there doesn't seem to be a way to change the color of the loaders. The number of loaders is a potential issue in most places and I'd like some feedback on how many you think I should use. Also, the colors are slightly different from the loaded version in just a couple of places, namely the NodeDetail panel and the table view when using skeleton loaders.

Let me know what you think. I'll add some screenshots below, but the deploy preview should give you a good example of how it will work in practice.

One final thing to consider, is what to do if the loading fails. Currently, it will just show skeleton loaders forever, which seems like an issue (aside from in the sidebar, where when it fails it will show nothing).

Workspaces sidebar:
image

Workspace overview:
image

Network overview:
image

Table overview:
image

Node overview (note the missing color from the large, full-width element on this screen):
image

### DIFF
--- a/src/components/NetworkPanel.vue
+++ b/src/components/NetworkPanel.vue
@@ -101,13 +101,19 @@
         </v-hover>
       </v-list-item-group>
     </template>
+
     <div
-      v-else
+      v-if="items.length === 0 && !loading"
       class="ws-detail-empty-list"
     >
       <v-icon color="blue lighten-1">
         info
       </v-icon> There's nothing here yet...
+    </div>
+
+    <div v-if="loading">
+      <v-skeleton-loader type="list-item" />
+      <v-skeleton-loader type="list-item" />
     </div>
   </v-list>
 </template>
@@ -147,6 +153,11 @@ export default Vue.extend({
 
     edgeTables: {
       type: Array as PropType<string[]>,
+      required: true,
+    },
+
+    loading: {
+      type: Boolean as PropType<boolean>,
       required: true,
     },
   },
--- a/src/components/NetworkPanel.vue
+++ b/src/components/NetworkPanel.vue
@@ -38,7 +38,21 @@
 
     <v-divider />
 
-    <template v-if="items.length > 0">
+    <div v-if="loading">
+      <v-skeleton-loader type="list-item" />
+      <v-skeleton-loader type="list-item" />
+    </div>
+
+    <div
+      v-else-if="items.length === 0"
+      class="ws-detail-empty-list"
+    >
+      <v-icon color="blue lighten-1">
+        info
+      </v-icon> There's nothing here yet...
+    </div>
+
+    <template v-else>
       <v-list-item-group color="primary">
         <v-hover
           v-for="item in items"
@@ -101,20 +115,6 @@
         </v-hover>
       </v-list-item-group>
     </template>
-
-    <div
-      v-if="items.length === 0 && !loading"
-      class="ws-detail-empty-list"
-    >
-      <v-icon color="blue lighten-1">
-        info
-      </v-icon> There's nothing here yet...
-    </div>
-
-    <div v-if="loading">
-      <v-skeleton-loader type="list-item" />
-      <v-skeleton-loader type="list-item" />
-    </div>
   </v-list>
 </template>
 
--- a/src/components/Sidebar.vue
+++ b/src/components/Sidebar.vue
@@ -107,6 +107,18 @@
             </v-list-item>
           </v-hover>
         </v-list-item-group>
+
+        <div v-if="loading">
+          <v-skeleton-loader
+            type="list-item"
+          />
+          <v-skeleton-loader
+            type="list-item"
+          />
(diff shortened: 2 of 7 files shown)


jtomeck (changes requested):
This is awesome @JackWilb. Only thing I'm noticing is some of the skeletons are rounded and some are square. I'm sure it's not intentional and I'm not sure what's causing it off the top of my head, but we should be consistent.

JackWilb:
> This is awesome @JackWilb. Only thing I'm noticing is some of the skeletons are rounded and some are square. I'm sure it's not intentional and I'm not sure what's causing it off the top of my head, but we should be consistent.

It was intentional :joy:. I had made the square ones because I had noticed that those list items (when loaded) didn't have rounded corners. Looking at it now, it appears that they all aren't rounded. 

Even still, my preference would be to use the rounded versions when it's loading, but I'm open to your feedback on how that looks and feels, Jared.

jtomeck:
> It was intentional

Haha, my bad. I assumed it was just in that one spot there was something causing it to happen.

> Even still, my preference would be to use the rounded versions when it's loading, but I'm open to your feedback on how that looks and feels, Jared.

I agree, let's do the rounded versions

JackWilb:
> I agree, let's do the rounded versions

Updated in 38b1061 and I updated the images above

jtomeck (approved):

waxlamp (approved):
Looks good. Do these animate? If so, can you post an mp4 of what that looks like?

The only thing I'd suggest adding (now or later) is to put in throbbers for the node and link count values. Those snap from 0 to the true value and as such they qualify for their own "skeleton" loaders as well.

JackWilb:
> Looks good. Do these animate? If so, can you post an mp4 of what that looks like?

https://user-images.githubusercontent.com/36867477/110147989-82a12800-7d99-11eb-849e-e1aac9e0fb08.mp4

> The only thing I'd suggest adding (now or later) is to put in throbbers for the node and link count values. Those snap from 0 to the true value and as such they qualify for their own "skeleton" loaders as well.

Added in 8036028. Here's a picture of how it looks:

image

waxlamp:
Thanks for the video preview.

For the network node/edge count, skeletons are not appropriate. Can you switch those out for throbbers or spinners instead? It should still show "nodes:" and "edges:" as in the final display; but while the count is being computed, the number should show up as a spinner.

Feel free to back revert the skeleton change and leave this task as a future issue; I suggested it now because it goes with the theme of filling in a proper empty state for the app.

JackWilb:
> Feel free to back revert the skeleton change and leave this task as a future issue; I suggested it now because it goes with the theme of filling in a proper empty state for the app.

See https://github.com/multinet-app/multinet-client/issues/157

src/components/NetworkPanel.vue
```diff
@@ -101,14 +101,20 @@
...
     >
       <v-icon color="blue lighten-1">
         info
       </v-icon> There's nothing here yet...
     </div>
+
+    <div v-if="loading">
+      <v-skeleton-loader type="list-item" />
+      <v-skeleton-loader type="list-item" />
+    </div>
```
waxlamp, line 117: It seems that we should restructure these `v-if`s a bit. Two thoughts here: Vue 2.1.0 has [v-else-if](https://vuejs.org/v2/guide/conditional.html#v-else-if); and we should be able to combine the changed clauses in this section with the big `v-if`ed `<template>` section above.

The overriding case is if we are in a loading state, so the `v-if=loading` case ought to go first. That can be followed up with a `v-else-if=items.length === 0` with the "there's nothing here yet" content, and then a `v-else` that has the big `<template>` section.

This would make this entire section much easier to understand. Let me know what you think.

waxlamp (changes requested):

src/components/NetworkPanel.vue
JackWilb: Good catch. I didn't think `v-else-if` was possible. I've cleaned it up in e1aaabd

waxlamp: Ah, looks great!

waxlamp (commented):
This looks good to me, but I haven't done a deep dive into the code--I'd wait on @AlmightyYakob's review before merging.

AlmightyYakob (changes requested):
Looks good just have a suggestion

AlmightyYakob (approved):
LGTM

jtomeck (approved):

waxlamp (approved):

PR #156, opened 2021-03-06
Add permissions check for creating networks and tables, disable menus

### LINKED ISSUE #83: Update app behavior to reflect user permissions
Including but not limited to:
- [ ] workspace listing
- [ ] action buttons
- [ ] AQL query allowances

JackWilb:
Closes #83

Adds a permissions check for the buttons on the network panel, and disables the add table and add network button, if the user is not a writer of the workspace.

Signed in:
image

Signed out:
image

### DIFF
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -9,6 +9,7 @@
         color="blue darken-2"
         icon
         medium
+        :disabled="!hasPerms"
         v-on="on"
       >
         <v-icon dark>
@@ -174,12 +175,14 @@
 
 <script lang="ts">
 import {
-  defineComponent, ref, Ref, computed,
+  defineComponent, ref, Ref, computed, watchEffect,
 } from '@vue/composition-api';
 
 import api from '@/api';
 import { TableFileType, CSVColumnType } from '@/types';
 import { validFileType, fileName as getFileName, analyzeCSV } from '@/utils/files';
+import { canUpload } from '@/utils/permissions';
+import store from '@/store';
 
 const defaultKeyField = '_key';
 const multinetTypes: readonly CSVColumnType[] = ['label', 'boolean', 'category', 'number', 'date'];
@@ -202,6 +205,13 @@ export default defineComponent({
   },
   setup(props, { emit }) {
     const { workspace } = props;
+
+    // Check for permissions
+    const permissions = computed(async () => api.getWorkspacePermissions(workspace));
+    const hasPerms = ref(false);
+    watchEffect(async () => {
+      hasPerms.value = canUpload(store.state.userInfo, await permissions.value);
+    });
 
     // Stepper control
     const step: Ref<number> = ref(1);
@@ -321,6 +331,7 @@ export default defineComponent({
     }
 
     return {
+      hasPerms,
       step,
       dialogWidth,
       headers,
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -207,10 +207,10 @@ export default defineComponent({
     const { workspace } = props;
 
     // Check for permissions
-    const permissions = computed(async () => api.getWorkspacePermissions(workspace));
     const hasPerms = ref(false);
     watchEffect(async () => {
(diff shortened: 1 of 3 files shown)


src/components/GraphDialog.vue
```diff
@@ -99,7 +104,24 @@ export default Vue.extend({
...
+  },
+  asyncComputed: {
+    permissions: {
+      async get(): Promise<WorkspacePermissionsSpec | null> {
+        try {
+          return api.getWorkspacePermissions(this.workspace);
+        } catch (error) {
+          return null;
+        }
+      },
```
AlmightyYakob, line 121: You only need `async` if you plan on calling `await` in the function
```suggestion
      get(): Promise<WorkspacePermissionsSpec | null> {
        try {
          return api.getWorkspacePermissions(this.workspace);
        } catch (error) {
          return null;
        }
      },
```

AlmightyYakob (changes requested):

src/components/GraphDialog.vue
JackWilb: I had copied the code from here for the async computed, and you'll see it has the same issue. Since this is in an async computed block, it requires that a promise is returned. That means that if we want to remove that async keyword, we'd have to replace 

```
        } catch (error) {
          return null;
        }
```

with

```
        } catch (error) {
          return new Promise(() => null);
        }
```

Or something similar.

src/components/TableDialog.vue
```diff
@@ -200,6 +203,15 @@ export default defineComponent({
...
   },
   setup(props, { emit }) {
+    const { workspace } = props;
+
+    // Check for permissions
+    const permissions = computed(async () => api.getWorkspacePermissions(workspace));
+    const hasPerms = ref(false);
+    watchEffect(async () => {
+      hasPerms.value = canUpload(store.state.userInfo, await permissions.value);
+    });
```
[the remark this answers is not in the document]
JackWilb: Thanks for the check here. I knew I was close, but it helps to have another set of eyes on it.

See 57b16ee

src/components/GraphDialog.vue
JackWilb: Do you think that's preferable to the current syntax?

AlmightyYakob (dismissed):

AlmightyYakob:
Actually hold off merging this, I'm still seeing some bugs

src/components/TableDialog.vue
```diff
@@ -200,6 +203,15 @@ export default defineComponent({
...
   },
   setup(props, { emit }) {
+    const { workspace } = props;
+
+    // Check for permissions
+    const hasPerms = ref(false);
+    watchEffect(async () => {
+      const perms = await api.getWorkspacePermissions(workspace);
+      hasPerms.value = canUpload(store.state.userInfo, perms);
+    });
```
AlmightyYakob, line 213: I believe there's a bug here, due to destructuring the `workspace` props. This kills the reactivity, so when you switch workspaces, this isn't updated. You should do the following instead:

```suggestion
    // Check for permissions
    const hasPerms = ref(false);
    watchEffect(async () => {
      const perms = await api.getWorkspacePermissions(props.workspace);
      hasPerms.value = canUpload(store.state.userInfo, perms);
    });
```

waxlamp (commented):
This looks ok to me. I'm deferring to @AlmightyYakob on a binding review for this PR. Thanks for doing this, @JackWilb!

waxlamp (commented):
What's going on with that empty app_logo.png file?

JackWilb:
> What's going on with that empty app_logo.png file?

I converted the ico to a png a while ago for something and I guess I forgot to remove it from the repo. I'll rebase it out of this PR, it shouldn't be here.

JackWilb:
This is on hold while I make the RFC for getting a users permissions for a workspace.

waxlamp:
> This is on hold while I make the RFC for getting a users permissions for a workspace.
> 
> See [multinet-app/multinet-rfcs#27](https://github.com/multinet-app/multinet-rfcs/pull/27)

I've switched this back to draft while we're working out the RFC and other implementation details.

PR #173, opened 2021-07-14
Allow for `import` statement in dependent libraries

waxlamp:
Building the client shows an error about `import` being a reserved
keyword in the generated `multinetjs` library. Setting `sourceType` to
`"module"` fixes the problem.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,6 +26,7 @@ module.exports = {
 
   parserOptions: {
     parser: '@typescript-eslint/parser',
+    sourceType: 'module',
     ecmaVersion: 2020,
     ecmaFeatures: {
       modules: true,


AlmightyYakob:
I actually don't think this PR solves the problem. I've dealt with this before when using `yarn link`, and I think the simple action of restarting the dev server after the initial error causes it to disappear. In my local testing I'm able to both make this error appear on this branch, as well as make it disappear on master. I don't know if there's any specific merits to setting `sourceType: 'module'`, but I just want to avoid confusion and make sure the problem and it's solution (or lack thereof) are clear. For that reason I'd suggest we close this PR, unless there are other reasons to change this setting.

waxlamp:
Ok, let's close this. But we'll need to rebase @eagw's checkout on `girder4` and ensure that she is not having the issue I was attempting to solve here.

PR #176, opened 2021-07-29
Fix workspace and table handling

eagw:
This PR changes how the responses from the server are handled to access table and workspace names properly.

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -76,7 +76,7 @@ const {
     async fetchWorkspaces(context) {
       const { commit } = rootActionContext(context);
       const workspaces = await api.workspaces();
-      commit.setWorkspaces(workspaces.results.map((h) => h.name));
+      commit.setWorkspaces(workspaces.results.map((w) => w.name));
     },
 
     async fetchWorkspace(context, workspace: string) {
@@ -90,9 +90,9 @@ const {
 
       commit.setCurrentWorkspace({
         name: workspace,
-        nodeTables: nodeTables.map((h) => h.name),
-        edgeTables: edgeTables.map((h) => h.name),
-        graphs: graphs.results.map((h) => h.name),
+        nodeTables: nodeTables.map((table) => table.name),
+        edgeTables: edgeTables.map((table) => table.name),
+        graphs: graphs.results.map((graph) => graph.name),
       });
     },
 
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -89,7 +89,10 @@ const {
       const edgeTables = tables.filter((table) => table.edge === true);
 
       commit.setCurrentWorkspace({
-        name: workspace, nodeTables: nodeTables.map((h) => h.name), edgeTables: edgeTables.map((h) => h.name), graphs: graphs.results.map((h) => h.name),
+        name: workspace,
+        nodeTables: nodeTables.map((h) => h.name),
+        edgeTables: edgeTables.map((h) => h.name),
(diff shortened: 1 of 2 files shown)


src/store/index.ts
```diff
@@ -76,27 +76,36 @@ const {
...
-      const nodeTables = await api.tables(workspace, { type: 'node' });
-      const edgeTables = await api.tables(workspace, { type: 'edge' });
       const graphs = await api.graphs(workspace);
+      const tables = (await api.tables(workspace)).results;
+      const nodeTables = tables.filter((table) => table.edge === false);
+      const edgeTables = tables.filter((table) => table.edge === true);
 
       commit.setCurrentWorkspace({
-        name: workspace, nodeTables, edgeTables, graphs,
+        name: workspace, nodeTables: nodeTables.map((h) => h.name), edgeTables: edgeTables.map((h) => h.name), graphs: graphs.results.map((h) => h.name),
```
waxlamp, line 92: You'll want to put each key/value pair on its own line to improve readability here.

src/views/TableDetail.vue
```diff
@@ -242,10 +242,12 @@ export default Vue.extend({
       });
 
       const {
-        rows,
+        results,
         count,
       } = result;
 
+      const rows = results;
+
```
waxlamp, line 250: I *think* the following is a more concise way to pull out the `results` property and rename it `rows` in one shot. You can try this out locally and see if it works.

```suggestion
        results: rows,
        count,
      } = result;

```

waxlamp (changes requested):
Looks pretty good; just a couple of suggestions.

src/store/index.ts
```diff
@@ -76,27 +76,36 @@ const {
     async fetchWorkspaces(context) {
       const { commit } = rootActionContext(context);
       const workspaces = await api.workspaces();
-      commit.setWorkspaces(workspaces);
+      commit.setWorkspaces(workspaces.results.map((h) => h.name));
```
AlmightyYakob, line 79: I think ideally we'd unwrap the paginated structure of the response, so that `.results` isn't needed, but I'm okay addressing that at a later time.

AlmightyYakob (approved):
LGTM!

src/store/index.ts
AlmightyYakob: Also, I'm not sure why `h` is used here. While this is very easy to follow, ideally something representative of the data would be used for a variable name (so here, it could be something as simple as `w`, for "workspace").

waxlamp: > Also, I'm not sure why `h` is used here. While this is very easy to follow, ideally something representative of the data would be used for a variable name (so here, it could be something as simple as `w`, for "workspace").

Or just `x` so that you don't have to necessarily pick a new letter for every time you use this pattern. But that being said, I think `h` is fine too since the intended semantics is pretty clear.

waxlamp: > I think ideally we'd unwrap the paginated structure of the response, so that `.results` isn't needed, but I'm okay addressing that at a later time.

Could you file an issue? This will definitely need to be addressed as soon as we ever have more than one page worth of workspaces.

src/views/TableDetail.vue
eagw: I tested it and it worked!

src/store/index.ts
eagw: See e4b6678

waxlamp (approved):

PR #178, opened 2021-08-05
Finish porting client to new backend

### LINKED ISSUE #161: Port client code to use girder 4 multinetjs
Depends on multinet-app/multinetjs#29.

eagw:
This PR changes how the client interacts with the data provided by multinetjs to display information about networks and nodes properly.  It also includes the following changes to adjust to the new backend:
- displaying user's initials
- changing to the appropriate localhost server address
- adjusting how table-network dependencies are calculated to warn users when they delete tables
- checking that tables aren't empty before attempting to display their rows

This PR works in conjunction with changes found in the [new-return-types](https://github.com/multinet-app/multinetjs/tree/new-return-types) branch of multinetjs

Resolves #161

### DIFF
--- a/src/components/DeleteTableDialog.vue
+++ b/src/components/DeleteTableDialog.vue
@@ -203,13 +203,18 @@ export default Vue.extend({
         });
 
         // eslint-disable-next-line no-underscore-dangle
-        const prelimNodes = nodes.results.map((node) => node._id.substring(0, node._id.indexOf('/')));
-        const nodeTables = prelimNodes.filter((node, index) => prelimNodes.indexOf(node) === index);
+        const prelimNodes = nodes.results.map((node) => node._id.split('/')[0]);
+        const nodeTables = [];
+        prelimNodes.forEach((table) => {
+          if (!nodeTables.includes(table)) {
+            nodeTables.push(table);
+          }
+        });
         // eslint-disable-next-line no-underscore-dangle
-        const edgeTable = edges.results.length > 0 ? edges.results[0]._id.substring(0, edges.results[0]._id.indexOf('/')) : '';
+        const edgeTable = edges.results.length > 0 ? edges.results[0]._id.split('/')[0] : '';
         const tables: string[] = [];
         selection.forEach((table) => {
-          if (table === edgeTable || nodeTables.indexOf(table) > -1) {
+          if (table === edgeTable || nodeTables.includes(table)) {
             tables.push(table);
           }
         });
--- a/src/components/DeleteTableDialog.vue
+++ b/src/components/DeleteTableDialog.vue
@@ -193,15 +193,23 @@ export default Vue.extend({
         workspace,
       } = this;
 
-      const graphNames = await api.graphs(workspace);
-
+      const graphNames = (await api.graphs(workspace)).results.map((graph) => graph.name);
       const using = [] as Array<{graph: string; tables: string[]}>;
       graphNames.forEach(async (graph) => {
-        const data = await api.graph(workspace, graph);
+        const nodes = await api.nodes(workspace, graph, {});
+        // eslint-disable-next-line no-underscore-dangle
+        const edges = await api.edges(workspace, graph, nodes.results[0]._id, {
+          direction: 'all',
+        });
 
+        // eslint-disable-next-line no-underscore-dangle
+        const prelimNodes = nodes.results.map((node) => node._id.substring(0, node._id.indexOf('/')));
+        const nodeTables = prelimNodes.filter((node, index) => prelimNodes.indexOf(node) === index);
+        // eslint-disable-next-line no-underscore-dangle
+        const edgeTable = edges.results.length > 0 ? edges.results[0]._id.substring(0, edges.results[0]._id.indexOf('/')) : '';
         const tables: string[] = [];
         selection.forEach((table) => {
-          if (table === data.edgeTable || data.nodeTables.indexOf(table) > -1) {
+          if (table === edgeTable || nodeTables.indexOf(table) > -1) {
             tables.push(table);
           }
         });
--- a/src/components/DeleteTableDialog.vue
+++ b/src/components/DeleteTableDialog.vue
@@ -194,9 +194,9 @@ export default Vue.extend({
         workspace,
       } = this;
 
-      function tableName(table: TableRow) {
+      function tableName(tableRow: TableRow) {
         // eslint-disable-next-line no-underscore-dangle
-        return table._id.split('/')[0];
+        return tableRow._id.split('/')[0];
       }
 
       const graphNames = (await api.graphs(workspace)).results.map((graph) => graph.name);
--- a/src/components/DeleteTableDialog.vue
+++ b/src/components/DeleteTableDialog.vue
@@ -113,6 +113,7 @@
 
 <script lang="ts">
 import Vue, { PropType } from 'vue';
+import { TableRow } from 'multinet';
 
 import api from '@/api';
 import { randomPhrase } from '@/utils/randomPhrase';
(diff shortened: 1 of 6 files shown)


src/components/DeleteTableDialog.vue
```diff
@@ -193,15 +193,23 @@ export default Vue.extend({
...
       graphNames.forEach(async (graph) => {
-        const data = await api.graph(workspace, graph);
+        const nodes = await api.nodes(workspace, graph, {});
+        // eslint-disable-next-line no-underscore-dangle
+        const edges = await api.edges(workspace, graph, nodes.results[0]._id, {
+          direction: 'all',
+        });
 
+        // eslint-disable-next-line no-underscore-dangle
+        const prelimNodes = nodes.results.map((node) => node._id.substring(0, node._id.indexOf('/')));
```
waxlamp, line 206: I think you can accomplish the splitting on `/` better with the use of [`String.split()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split) instead of the combination of `substring()` and `indexOf()`.

src/components/DeleteTableDialog.vue
```diff
@@ -193,15 +193,23 @@ export default Vue.extend({
...
+        const nodeTables = prelimNodes.filter((node, index) => prelimNodes.indexOf(node) === index);
+        // eslint-disable-next-line no-underscore-dangle
+        const edgeTable = edges.results.length > 0 ? edges.results[0]._id.substring(0, edges.results[0]._id.indexOf('/')) : '';
```
waxlamp, line 209: Same thing about `String.split()` as above.

src/components/DeleteTableDialog.vue
```diff
@@ -193,15 +193,23 @@ export default Vue.extend({
...
         const tables: string[] = [];
         selection.forEach((table) => {
-          if (table === data.edgeTable || data.nodeTables.indexOf(table) > -1) {
+          if (table === edgeTable || nodeTables.indexOf(table) > -1) {
```
waxlamp, line 212: Check out the [`Array.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) function--it's probably better here than `indexOf()`.

src/views/GraphDetail.vue
```diff
@@ -427,19 +427,22 @@ export default Vue.extend({
...
-      const edgeQuery = `
-        FOR table in [${graph.edgeTable}]
-          RETURN LENGTH(table)
-      `;
-      const edges = await api.aql(this.workspace, edgeQuery);
+      // eslint-disable-next-line no-underscore-dangle
+      const prelimNodes = nodes.results.map((node) => node._id.substring(0, node._id.indexOf('/')));
+      this.nodeTypes = prelimNodes.filter((node, index) => prelimNodes.indexOf(node) === index);
+      // eslint-disable-next-line no-underscore-dangle
+      this.edgeTypes = edges.results.length > 0 ? [edges.results[0]._id.substring(0, edges.results[0]._id.indexOf('/'))] : [];
```
waxlamp, line 442: More possibilities for `split()` and `includes()` here.

waxlamp: Further, it seems like some of the ID splitting could perhaps be factored out into a function. It might make this type of logic more readable, or maybe not. Thoughts, @AlmightyYakob?

src/views/NodeDetail.vue
```diff
@@ -391,27 +391,34 @@ export default Vue.extend({
   methods: {
     async update() {
       this.loading = true;
-      const attributes = await api.attributes(this.workspace, this.graph, `${this.type}/${this.node}`);
-      const incoming = await api.edges(this.workspace, this.graph, `${this.type}/${this.node}`, {
+      const nodeTable = (await api.table(this.workspace, this.type, {})).results;
+      // eslint-disable-next-line no-underscore-dangle
+      const node = nodeTable[nodeTable.map((table) => table._key).indexOf(this.node)];
```
waxlamp, line 396: I'm almost certain you can use something like [`Array.find()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) here instead of this rather complex array indexing expression.

waxlamp (changes requested):
Code looks good overall. Made a few suggestions of cleaner ways to implement some stuff.

waxlamp:
@AlmightyYakob, I requested a review from you on this just now to get one more set of eyes on the overall logic here. I gave a lot of line-oriented feedback in my review and I think something more high level from you would be valuable here.

src/components/DeleteTableDialog.vue
eagw: See https://github.com/multinet-app/multinet-client/commit/ceedc898a945927e4e277d13583c9eec195e6c64#diff-59d96b3b10e2aff2a38e84b28bc951f257462a7772757d307d3952647372bd4dR206

src/components/DeleteTableDialog.vue
eagw: See https://github.com/multinet-app/multinet-client/commit/ceedc898a945927e4e277d13583c9eec195e6c64#diff-59d96b3b10e2aff2a38e84b28bc951f257462a7772757d307d3952647372bd4dR207-R212

To replace indexOf() I had to use a loop instead of filter() because filter() doesn't assign a value to nodeTables until it's iterated through the entire list.  Do you still prefer that to using indexOf() @waxlamp?

src/components/DeleteTableDialog.vue
eagw: See https://github.com/multinet-app/multinet-client/commit/ceedc898a945927e4e277d13583c9eec195e6c64#diff-59d96b3b10e2aff2a38e84b28bc951f257462a7772757d307d3952647372bd4dL212-R217

src/views/GraphDetail.vue
eagw: See https://github.com/multinet-app/multinet-client/commit/ceedc898a945927e4e277d13583c9eec195e6c64#diff-c702a8a412d06f504ab01fd49432140ee09ea842900696710d609086d3393813L439-R446 and my comment above about replacing indexOf() with includes()

src/views/NodeDetail.vue
eagw: See https://github.com/multinet-app/multinet-client/commit/ceedc898a945927e4e277d13583c9eec195e6c64#diff-24af31592a6444e4718ffd1207989976b07faaae1e61bc74c9da27b0537adbf9L396-R396

eagw:
One thing I forgot about is the issue with the async functions in [DeleteTableDialog.vue](https://github.com/multinet-app/multinet-client/pull/178/files#diff-59d96b3b10e2aff2a38e84b28bc951f257462a7772757d307d3952647372bd4d) that I briefly mentioned @waxlamp , but we never got to talk about.  From what I've tested, I don't think the warning feature about table dependencies worked previously and since what I changed is meant to be a direct replacement of that logic the feature still doesn't work currently.  I have figured out a way to solve the issue, but I wasn't sure if I should create an issue and start an additional pull request or add it here.

src/views/GraphDetail.vue
```diff
@@ -427,19 +427,26 @@ export default Vue.extend({
         offset: this.offset,
         limit: this.limit,
       });
+      // eslint-disable-next-line no-underscore-dangle
+      const edges = await api.edges(this.workspace, this.graph, nodes.results[0]._id, {
```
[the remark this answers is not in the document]
eagw: See 2cce969

src/components/DeleteTableDialog.vue
```diff
@@ -193,25 +194,27 @@ export default Vue.extend({
         workspace,
       } = this;
 
+      function tableName(table: TableRow) {
+        // eslint-disable-next-line no-underscore-dangle
+        return table._id.split('/')[0];
+      }
```
eagw, line 200: Is this what you two were thinking about @waxlamp and @AlmightyYakob ?

AlmightyYakob: Looks good to me :+1:

AlmightyYakob: Actually, I think `table` should be named `tableRow` or something, since it's not actually a table itself, it's a row of a table. Otherwise it looks good.

eagw: That makes sense.
See 2f51fba

waxlamp (commented):
If @AlmightyYakob is happy, then I'm happy, so please just go with his review feedback on this. Thanks for your efforts here @eagw!

waxlamp:
> One thing I forgot about is the issue with the async functions in [DeleteTableDialog.vue](https://github.com/multinet-app/multinet-client/pull/178/files#diff-59d96b3b10e2aff2a38e84b28bc951f257462a7772757d307d3952647372bd4d) that I briefly mentioned @waxlamp , but we never got to talk about. From what I've tested, I don't think the warning feature about table dependencies worked previously and since what I changed is meant to be a direct replacement of that logic the feature still doesn't work currently. I have figured out a way to solve the issue, but I wasn't sure if I should create an issue and start an additional pull request or add it here.

Since you've determined that the feature was already broken before, there's no need to fix it here. The best thing to do is file an issue with some background and any ideas you have for solving it. Then we can pick up that fix in a subsequent pull request.

PR #179, opened 2021-08-10
Support S3 File Field Upload functionality

eagw:
Update the client to pass additional information to multinetjs to allow for table and network uploads using the S3 File Field client.

This PR works in conjunction with the s3ff branch of [multinetjs](https://github.com/multinet-app/multinetjs/tree/s3ff)

Should be merged after #178.  Commits pertinent to this PR are 9e1d1df, 86dd4fc and fcaec43, which make changes to FileUploadForm.vue and TableDialog.vue (all other commits are for changes addressed in #178).
**Edit: Now that #178 is merged, all commits associated with this PR are relevant.**

### DIFF
--- a/src/components/FileUploadForm.vue
+++ b/src/components/FileUploadForm.vue
@@ -77,7 +77,6 @@ import Vue, { PropType } from 'vue';
 
 import api from '@/api';
 import { FileType } from '@/types';
-import { host } from '@/environment';
 
 export default Vue.extend({
   name: 'FileUploadForm',
@@ -184,7 +183,6 @@ export default Vue.extend({
         await api.uploadNetwork(workspace, fileName, {
           type: selectedType.queryCall as UploadType,
           data: file,
-          url: host,
         }, {
           onUploadProgress: this.handleUploadProgress,
         });
--- a/src/components/FileUploadForm.vue
+++ b/src/components/FileUploadForm.vue
@@ -77,6 +77,7 @@ import Vue, { PropType } from 'vue';
 
 import api from '@/api';
 import { FileType } from '@/types';
+import { host } from '@/environment';
 
 export default Vue.extend({
   name: 'FileUploadForm',
@@ -180,9 +181,10 @@ export default Vue.extend({
       this.uploading = true;
 
       try {
-        await api.uploadTable(workspace, fileName, {
+        await api.uploadNetwork(workspace, fileName, {
           type: selectedType.queryCall as UploadType,
           data: file,
+          url: host,
         }, {
           onUploadProgress: this.handleUploadProgress,
         });
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -218,7 +218,12 @@ export default defineComponent({
       }));
     });
 
-    const edgeTable = computed(() => Object.prototype.hasOwnProperty.call(sampleRows.value[0], '_from') && Object.prototype.hasOwnProperty.call(sampleRows.value[0], '_to'));
+    const edgeTable = computed(() => {
+      const sample = sampleRows.value[0] || {};
+      const hasFrom = Object.prototype.hasOwnProperty.call(sample, '_from');
+      const hasTo = Object.prototype.hasOwnProperty.call(sample, '_to');
+      return hasFrom && hasTo;
+    });
 
     // Type recommendation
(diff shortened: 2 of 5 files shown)


src/components/TableDialog.vue
```diff
@@ -218,6 +219,11 @@ export default defineComponent({
       }));
     });
 
+    const edgeTable = computed(() => {
+      const keys = Object.keys(sampleRows.value[0] || {});
+      return keys.includes('_from') && keys.includes('_to');
```
waxlamp, line 224: There's a better way to do this. Extract the object from `sampleRows` as you are doing in line 223, then call `.hasOwnProperty()` on it to detect the `_from` and `_to` attributes.

You may run into a linter error when you do this--this [documentation](https://eslint.org/docs/rules/no-prototype-builtins) explains what's going on and how to fix it.

src/components/FileUploadForm.vue
```diff
@@ -180,9 +181,10 @@ export default Vue.extend({
       this.uploading = true;
 
       try {
-        await api.uploadTable(workspace, fileName, {
+        await api.uploadNetwork(workspace, fileName, {
           type: selectedType.queryCall as UploadType,
           data: file,
+          url: host,
```
waxlamp, line 187: If my [comment](https://github.com/multinet-app/multinetjs/pull/44#pullrequestreview-728659778) on the corresponding multinetjs PR is correct, then this will need to change.

waxlamp (changes requested):

src/components/TableDialog.vue
```diff
@@ -295,9 +301,11 @@ export default defineComponent({
         await api.uploadTable(workspace, fileName.value, {
           type: 'csv',
           data: selectedFile.value,
+          edgeTable: edgeTable.value,
           key: keyField.value,
           overwrite: overwrite.value,
           columnTypes: columnType.value,
+          url: host,
```
AlmightyYakob, line 308: Same as roni's comment about the corresponding change in multinetJS

eagw: See 5f46644

src/components/TableDialog.vue
```diff
@@ -218,10 +218,7 @@ export default defineComponent({
       }));
     });
 
-    const edgeTable = computed(() => {
-      const keys = Object.keys(sampleRows.value[0] || {});
-      return keys.includes('_from') && keys.includes('_to');
-    });
+    const edgeTable = computed(() => Object.prototype.hasOwnProperty.call(sampleRows.value[0], '_from') && Object.prototype.hasOwnProperty.call(sampleRows.value[0], '_to'));
```
waxlamp, line 221: For readability, you may want to split this up into a couple of `const` expressions, like:

```typescript
const hasFrom = Object.prototype.hasOwnProperty.call(sampleRows.value[0]);
```

In general, when you have complex expressions composed into a long line of code, it's good to consider the structure of it for purposes of readability to other devs (including Emi-from-the-future).

I think you also want to extract the `keys` value as in your old code:

```typescript
const sample = sampleRows.value[0] || {};
```

so that you aren't operating on `undefined` by accident. Essentially, you want the type of `sample` to be `Node` rather than `Node | undefined` (I'm saying `Node` because I forgot the true name of the type, but my point stands).

src/components/TableDialog.vue
eagw: That's helpful, thank you!

See 701bec8

src/components/TableDialog.vue
eagw: Thank you for the feedback, those are both great points!

See a970386

src/components/FileUploadForm.vue
```diff
@@ -180,7 +180,7 @@ export default Vue.extend({
       this.uploading = true;
 
       try {
-        await api.uploadTable(workspace, fileName, {
+        await api.uploadNetwork(workspace, fileName, {
```
AlmightyYakob, line 183: Why was this changed to `uploadNetwork`?

waxlamp (commented):
LGTM, pending Jake's approval.

src/components/FileUploadForm.vue
eagw: This is the component that handles uploading files to create networks, so I switched it to use the new separate function.  TableDialog.vue takes care of uploading files to make tables.

AlmightyYakob (approved):

PR #184, opened 2021-08-13
Always run CI

AlmightyYakob:
Previously, CI would only run on PRs and pushes to `master`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,14 +1,6 @@
 name: CI
 
-on:
-  pull_request:
-    branches:
-      - master
-
-  push:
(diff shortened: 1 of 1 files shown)


waxlamp (approved):
Practically speaking, this just adds CI runs on PRs that are running against any branch (not just `master`), and on mere pushes to any branch (not just merges of PRs to `master`)--is that right?

AlmightyYakob:
> Practically speaking, this just adds CI runs on PRs that are running against any branch (not just `master`), and on mere pushes to any branch (not just merges of PRs to `master`)--is that right?

Yes exactly

PR #187, opened 2021-08-18
Refactor client to work with new workspace renaming endpoint

eagw:
Removes the use of a destructuring assignment for the data variable and extracts the new name of the workspace to account for the new Workspace return type of renameWorkspace() and support the use of the new workspace renaming endpoint.

Works in conjunction with the rename-workspace branches of [multinetjs](https://github.com/multinet-app/multinetjs/tree/rename-workspace) and [multinet-api](https://github.com/multinet-app/multinet-api/tree/rename-workspace).

### DIFF
--- a/src/views/WorkspaceDetail.vue
+++ b/src/views/WorkspaceDetail.vue
@@ -258,7 +258,7 @@ export default Vue.extend({
 
       if (this.localWorkspace !== null) {
         try {
-          const { data } = await api.renameWorkspace(this.workspace, this.localWorkspace);
+          const data = (await api.renameWorkspace(this.workspace, this.localWorkspace)).name;
           this.$router.push(`/workspaces/${data}`);
           this.editing = false;
           this.requestError = null;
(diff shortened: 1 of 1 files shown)


src/views/WorkspaceDetail.vue
```diff
@@ -258,7 +258,7 @@ export default Vue.extend({
 
       if (this.localWorkspace !== null) {
         try {
-          const { data } = await api.renameWorkspace(this.workspace, this.localWorkspace);
+          const data = (await api.renameWorkspace(this.workspace, this.localWorkspace)).name;
```
waxlamp, line 261: The following might be more readable (if it works correctly):

```suggestion
          const { name } = (await api.renameWorkspace(this.workspace, this.localWorkspace));
```

In order to use this, you would need to also change `data` to `name` in the following lines (so, rather than accepting this code suggestion, I would incorporate it into your own local edits and make sure it works as intended).

This is better because it reflects the actual usage of values in the variable names.

waxlamp (changes requested):
Looks good, just one readability suggestion.

src/views/WorkspaceDetail.vue
eagw: That works!

See d4ba127

waxlamp (approved):

AlmightyYakob (approved):

PR #188, opened 2021-08-18
Replace all use of 'graph(s)' with 'network(s)' for consistency

eagw:
Replaces all use of graph(s) with network(s) to maintain continuity across the frontend.  This includes the renaming of: 
- GraphDetail.vue
- DeleteGraphDialog.vue
- GraphCreateForm.vue
- GraphDialog.vue
- GraphUploadForm.vue

The only places where graph has not been replaced are in the urls which link to the visual applications (where a similar change has not been made).

Works in conjunction with the graph-to-network branch of [multinetjs](https://github.com/multinet-app/multinetjs/tree/graph-to-network).

### DIFF
--- a/src/views/NetworkDetail.vue
+++ b/src/views/NetworkDetail.vue
@@ -1,6 +1,6 @@
 <template>
   <v-container
-    class="graph-container"
+    class="network-container"
     fill-height
     fluid
     pa-0
@@ -34,7 +34,7 @@
             class="mr-3"
             color="grey lighten-1"
           >timeline</v-icon>
-          {{ graph }}
+          {{ network }}
         </span>
       </v-toolbar-title>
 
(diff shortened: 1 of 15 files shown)


src/views/NetworkDetail.vue
```diff
@@ -332,18 +332,18 @@ import api from '@/api';
...
 
 export default Vue.extend({
-  name: 'GraphDetail',
+  name: 'NetworkDetail',
   filters: {
     appendArgs(url: string) {
-      return `${url}/?workspace=${this.workspace}&graph=${this.graph}`;
+      return `${url}/?workspace=${this.workspace}&graph=${this.network}`;
     },
   },
```
AlmightyYakob, line 340: The `appendArgs` filter doesn't appear to be used anywhere, so I think we can just delete this entire `filters` block.

AlmightyYakob (approved):
One housekeeping suggestion, but otherwise lgtm

src/views/NetworkDetail.vue
eagw: See 88c3a3f

PR #196, opened 2021-10-01
Guess the column types from the first 1000 rows

### LINKED ISSUE #195: On upload the client is processing the entire dataset to determine types -- slow for large uploads
I'm pretty sure the client is processing the entirety of the data to determine the column type before upload. We should put some kind of limit on this to speed it up for the user so that they're not waiting for their browser to finish running our JS before being allowed to upload. I imagine this experience would be notably bad on slow hardware. Maybe, if it's over 1000 (this limit is completely arbitrary and could be much larger) rows, we just say "tell us your types" instead of making an assumption for them.

This affects all branches currently, but is particularly a problem on `girder4` since that branch the corresponding branch in the api enables large file uploads

JackWilb:
Closes #195

Instead of parsing the whole dataset to guess the types for the columns, we can just parse the first thousand rows. This helps the run time -- it's basically instant. That means the preview rows will have loaded by the time you're looking at type specifying page in the stepper.

I tried to use the papaparse `preview` [option](https://www.papaparse.com/docs#config), but I think the `step` callback overrides that. To get around that issue I implemented a counter that will trigger the parser to abort when we hit that number.

The types suggested were identical on my test set with either so I think this is pretty robust.

We'll need some better error UI in the future to alert users to any parsing issues, but this has never been implemented and shouldn't slow this PR down.

### DIFF
--- a/src/utils/files.ts
+++ b/src/utils/files.ts
@@ -56,16 +56,23 @@ async function analyzeCSV(file: File): Promise<CSVAnalysis> {
     const columnTypes = new Map<string, TypeScore>();
     const typeRecs = new Map<string, CSVColumnType>();
     const sampleRows = [] as Array<{}>;
+    let rowsParsed = 0;
 
     Papa.parse(file, {
       header: true,
       skipEmptyLines: true,
-      step(row: ParseResult<{}>) {
+      step(row: ParseResult<{}>, parser) {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         const data = row.data as { [key: string]: any };
 
         if (sampleRows.length !== 30) {
           sampleRows.push({ ...data });
+        }
+
+        // Increment number of rows parsed and return if we've done 1000
+        rowsParsed += 1;
+        if (rowsParsed === 1000) {
+          parser.abort();
         }
 
         Object.keys(data).forEach((key: string) => {


AlmightyYakob:
> I tried to use the papaparse `preview` [option](https://www.papaparse.com/docs#config), but I think the `step` callback overrides that. To get around that issue I implemented a counter that will trigger the parser to abort when we hit that number.

Since we're now using the `preview` option, I don't actually think we need to use the `step` function at all. Since the parsing result will only ever be at most 1000 rows, and `complete` takes the parsed results as its first argument (even though we currently ignore it), we could just do all that we're currently doing in the `step` function, in the `complete` function instead.

This would make the code slightly simpler, and may improve performance slightly, but in the end would accomplish the same thing. What you've done looks fine to me (and worked fine when tested locally), so I'll give an approval now. If you want to update this to implement what I've suggested, feel free to do that and just re-request my review.

AlmightyYakob (approved):

JackWilb:
> This would make the code slightly simpler, and may improve performance slightly, but in the end would accomplish the same thing. What you've done looks fine to me (and worked fine when tested locally), so I'll give an approval now. If you want to update this to implement what I've suggested, feel free to do that and just re-request my review.

I'll leave it as is for now, but I'm open to refactoring at some point

Issue #194, opened 2021-09-16
User initial don't display properly in girder4


naglepuff:
During the calculation of user initials in `LoginMenue.vue`, there is no fallback if `first_name` and `last_name` are `''`, `null` or `undefined`. This doesn't break anything, but we should have placeholder values for this case, and trying to squeeze `undefinedundfined` into the user icon looks bad.

waxlamp:
Not to get too philosophical, but if `undefinedundefined` makes it to the end user, I consider that a broken part of the client.

JackWilb:
Closed via #204

Issue #190, opened 2021-08-19
Text in Delete Workspace Dialog Persists


naglepuff:
I am able to see this small quirk on [multinet.app](multinet.app) and [web.next.multinet.app](web.next.multinet.app).

To reproduce:

1. Go to one of the websites mentioned above
2. Click on the "Delete" icon in the workspace sidebar
3. Enter some text when it asks for a random phrase, then cancel
4. Open up the dialog again by clicking a "Delete" icon
5. The text you just entered remains

I am of the opinion that this text should be cleared when the delete dialog is closed. This might also affect table/network deletion. Not a huge deal, and as far as I can tell it doesn't really impact any functionality, more of a quality-of-life thing.

waxlamp:
I agree that this needs to be fixed. I'm marking it as "good first issue" but feel free to fix it at your own leisure as well.

JackWilb:
Closed via #204

PR #204, opened 2021-11-17
Address various papercuts in the UI/UX

JackWilb:
As I've been using the app I've noticed many small UX issues that make the app harder to use. This is a first attempt at addressing some of the more heinous ones:

- #194 Handle user initals better if they don't exist
- #190 Clear deletion confirmation text when the require phrase changes and on successful delete
- Sort all networks, tables, and workspaces alphabetically
- Only show skeletons when they're needed in table detail
- Use the same menu with permissions and AQL editor wherever the menu icon appears
- Disable the create network and table buttons when a user has clicked them (I've seen users click multiple time creating multiple upload objects)
- Fix up the logic for file upload handling so it clears properly when a file is removed/added. This change improves the code readability
- Make the error messages on workspace rename and table upload failure more verbose

### DIFF
--- a/src/components/DeleteNetworkDialog.vue
+++ b/src/components/DeleteNetworkDialog.vue
@@ -134,8 +134,6 @@ export default Vue.extend({
 
       await Promise.all(selection.map((network) => api.deleteNetwork(workspace, network)));
       this.dialog = false;
-
-      this.clear();
     },
 
     clear() {
--- a/src/components/DeleteNetworkDialog.vue
+++ b/src/components/DeleteNetworkDialog.vue
@@ -117,6 +117,7 @@ export default Vue.extend({
   watch: {
     dialog() {
       if (this.dialog) {
+        this.clear();
         this.confirmationPhrase = randomPhrase();
       } else {
         this.$emit('closed');
@@ -133,6 +134,13 @@ export default Vue.extend({
 
       await Promise.all(selection.map((network) => api.deleteNetwork(workspace, network)));
       this.dialog = false;
+
+      this.clear();
+    },
+
+    clear() {
+      this.confirmationPhrase = '';
+      this.confirmation = '';
     },
   },
 });
--- a/src/components/TableDialog.vue
+++ b/src/components/TableDialog.vue
@@ -163,6 +163,8 @@
             <v-btn
               color="primary"
               depressed
+              :loading="loading"
+              :disabled="loading"
               @click="createTable"
             >
               Create Table
@@ -275,6 +277,7 @@ export default defineComponent({
     const tableDialog = ref(false);
     const tableCreationError = ref<string | null>(null);
     const createDisabled = computed(() => selectedFile.value === null || !fileName.value);
+    const loading = ref(false);
     async function createTable() {
       if (selectedFile.value === null || fileName.value === null) {
         return;
@@ -284,6 +287,7 @@ export default defineComponent({
       uploading.value = true;
 
       try {
+        loading.value = true;
         await api.uploadTable(workspace, fileName.value, {
           data: selectedFile.value,
           edgeTable: edgeTable.value,
@@ -292,6 +296,7 @@ export default defineComponent({
 
         tableCreationError.value = null;
         tableDialog.value = false;
+        loading.value = false;
 
         emit('success');
         resetAllFields();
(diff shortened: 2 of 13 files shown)


src/store/index.ts
```diff
@@ -37,23 +37,30 @@ const {
     currentWorkspacePermission: null,
   } as State,
   getters: {
+    tables(state: State, getters): string[] {
+      if (state.currentWorkspace !== null && state.currentWorkspace.nodeTables && state.currentWorkspace.edgeTables) {
+        return getters.nodeTables.concat(getters.edgeTables).sort();
+      }
+      return [];
```
waxlamp, line 44: What problem does it solve to alphabetize all the tables? There's an argument that the tables should appear in order of newness (though I'm not sure that's what happens currently), but alphabetic is a fine default choice too.

More generally, we ought to be able to change the sort order on the fly. Would you mind filing an issue to that effect?

src/views/TableDetail.vue
```diff
@@ -18,7 +18,7 @@
 
           <v-divider />
 
-          <div v-if="loading">
+          <div v-if="loadingTables">
```
waxlamp, line 21: Is `loading` no longer used in this component anywhere?

src/components/DeleteNetworkDialog.vue
```diff
@@ -117,6 +117,7 @@ export default Vue.extend({
   watch: {
     dialog() {
       if (this.dialog) {
+        this.clear();
```
waxlamp, line 120: Is there a reason the clear needs to be done here and after closing the dialog (i.e., down on line 138)?

src/components/TableDialog.vue
```diff
@@ -45,15 +45,16 @@
                 <v-layout wrap>
                   <v-flex>
                     <v-file-input
-                      :error-messages="fileUploadError"
+                      v-model="selectedFile"
```
waxlamp, line 50: Could you explain what the changes in this commit accomplish? It's hard to tell from reading the code. In particular, it looks like the error message bit was stripped out and I'm not sure if it's been shifted to somewhere else. But overall I'd like to understand better what this does.

src/views/WorkspaceDetail.vue
```diff
@@ -228,7 +228,8 @@ export default Vue.extend({
           if (err.response.status === 409) {
             this.requestError = 'A workspace by that name already exists';
           } else {
-            this.requestError = err.response.statusText;
+            console.log(err.response);
```
waxlamp, line 231: Could you get rid of this `console.log()` statement?

waxlamp (approved):
Looks like a slew of useful improvements, thanks for doing this @JackWilb. I just had a few questions and one request for removing a console.log, but I'm marking this as approved so you can merge without a second round of reviews.

src/views/WorkspaceDetail.vue
JackWilb: See 8d68860. I've also opened #206 to stop us (mostly me, lol) from committing code that doesn't pass the production linter

src/views/TableDetail.vue
JackWilb: It's used for the rows in the table when the table changes. The problem was that when you were in the view and clicked on another table, you'd have the skeleton loaders pop back in for the table list, even though the list of tables was loaded and still visible under the skeletons.

src/components/TableDialog.vue
JackWilb: Yes, it appeared to me that we were rolling our own validation and logic to handle file names, when that could be more easily handled by the vuetify component. 

The main goal of the commit was to update the table name when files were changed/removed so that you didn't upload a csv called "people.csv" to a table called "connections" (which I have done 🙃).

When I started making that change, I saw we should be using the `accept` and `v-model` attributes of the vuetify components,. which led to removing a ton of custom logic. For example, the table uploader should only allow .csv, then we don't need a custom function to check if the file is a csv; it's guaranteed that it will be a csv by the `accept` property.

waxlamp: Great, thanks Jack!

src/components/DeleteNetworkDialog.vue
JackWilb: I've updated this to just clear here (where you've commented), which is when the message is reset. Good catch

I did this across all the deletion components in e3a1b89

src/store/index.ts
waxlamp: I was still curious why you wanted to see tables alphabetized.

PR #209, opened 2021-11-19
Use the organization action for linting

JackWilb:
Nope, it's just an improvement to help us reduce redundant code across the repos.

I've added an action in multinet-app/.github that handles this logic. The gameplan is to prefer using that to workflow files in each individual repo to provide more consistent testing logic.

N/A

TODOs:
- [x] Make sure it works
- [x] Change the branch protection rules to make it use the new new for the workflow

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,34 +1,16 @@
-name: CI
+name: Lint application using multinet-app/.github generic lint workflow
 
-on: [push, pull_request]
+on: 
+  pull_request:
+    branches:
+      - '**'
+  
+  push:
+    branches:
+    - main
+
 jobs:
-  build-and-test:
-    name: Build and test on Ubuntu - Node 10
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@master
-
-      # Copy Env File
-      - run: cp .env.default .env
-
-      # Build and test Multinet client app.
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
I'm confused by this. Why place the action in a separate repo?

JackWilb:
> I'm confused by this. Why place the action in a separate repo?

It allows us to reduce redundant actions across the repos. Instead, it's now in one central place and can be reused. I'm also using it for PR templates on all our repos

AlmightyYakob:
> It allows us to reduce redundant actions across the repos. Instead, it's now in one central place and can be reused. I'm also using it for PR templates on all our repos

I see, I was slightly confused. I think this is fine, although one thing I'll say is I don't think the app registry step should live in that lint action, since it only applies to this repo. Can't that step just be placed here, before the use of that action?

JackWilb:
> I see, I was slightly confused. I think this is fine, although one thing I'll say is I don't think the app registry step should live in that lint action, since it only applies to this repo. Can't that step just be placed here, before the use of that action?

That's not possible because the checkout and yarn steps all happen in the other action. If we wanted that, we'd have to do the checkout here, and then the `yarn init-appregistry`. I think it's okay to just pass through a variable that triggers it if it needs it, since it's just a one off. If that changes, we can update it pretty easily

PR #210, opened 2021-12-10
Show upload status in workspace detail view

### LINKED ISSUE #174: Show upload status in UI
Since uploads are now done asynchronously, we need a way to show the user what the status of these uploads is in the workspace detail view, etc.

JackWilb:
Closes #174

This PR does 3 things:
1. Converts the workspace detail component to composition API (there's one small wrinkle with vue.router that I had to work around. See [here](https://github.com/multinet-app/multinet-client/pull/210/commits/4d45316edcaf1c3f61519487604ffb8b33058496#diff-800e682df700e57321ffeca2830a5d4fb044ce70407890f05188386a94ac54c2R136-R138)) (See code review for updated approach)
2. Gets all uploads from the API using a new route added in multinetjs
3. Adds an info alert to the workspace detail view when there are uploads that are not finished

With upload in progress:
image

Without upload in progress (same as before):
image

TODOs:
- [x] Make a PR for the multinetjs change (multinet-app/multinetjs#66)
- [x] Bump multinetjs version here
- [x] Add an animation
- [x] Figure out what upload statuses there are (what should we do on failed, etc.?) and update the logic to filter them (#211)

### DIFF
--- a/src/views/WorkspaceDetail.vue
+++ b/src/views/WorkspaceDetail.vue
@@ -97,7 +97,18 @@
             type="info"
             class="mb-0"
           >
-            Uploading: {{ upload.blob.substring(upload.blob.indexOf('/') + 1) }}
+            <v-row align="center">
+              <v-col class="grow">
+                Uploading: {{ upload.blob.substring(upload.blob.indexOf('/') + 1) }}
+              </v-col>
+              <v-col class="shrink">
+                <v-progress-circular
+                  indeterminate
+                  color="white"
+                  size="26"
+                />
+              </v-col>
+            </v-row>
           </v-alert>
         </v-col>
       </v-row>
--- a/src/views/WorkspaceDetail.vue
+++ b/src/views/WorkspaceDetail.vue
@@ -129,7 +129,13 @@
 </template>
 
 <script lang="ts">
-import Vue, { PropType } from 'vue';
+import {
+  defineComponent, ref, PropType, computed, watch,
+} from '@vue/composition-api';
+
+// import { useRouter } from 'vue-router'; // Router import vue3 syntax
+// eslint-disable-next-line import/no-cycle
+import router from '@/router'; // Temporary router syntax that lets us use the old push
 
 import api from '@/api';
 import TablePanel from '@/components/TablePanel.vue';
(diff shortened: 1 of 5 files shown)


src/views/WorkspaceDetail.vue
```diff
@@ -143,107 +163,109 @@ const workspaceNameRules: Array<(x: string) => string|boolean> = [
...
-    };
-  },
 
-  computed: {
-    nodeTables: () => store.getters.nodeTables,
-    edgeTables: () => store.getters.edgeTables,
-    tables: () => store.getters.tables,
-    networks: () => store.getters.networks,
+  setup(props) {
+    // const router = useRouter();
```
AlmightyYakob, line 183: This is how I use the router with CompositionAPI on Vue2

```suggestion
  setup(props, ctx) {
    const router = ctx.root.$router;
```

src/views/WorkspaceDetail.vue
```diff
@@ -122,14 +136,20 @@
...
 
 <script lang="ts">
-import Vue, { PropType } from 'vue';
+import {
+  defineComponent, ref, PropType, computed, watch,
+} from '@vue/composition-api';
+
+// import { useRouter } from 'vue-router'; // Router import vue3 syntax
+// eslint-disable-next-line import/no-cycle
+import router from '@/router'; // Temporary router syntax that lets us use the old push
```
AlmightyYakob, line 152:
```suggestion
```

If my suggestion around `setup` is committed, this is no longer needed.

src/views/WorkspaceDetail.vue
```diff
@@ -143,107 +163,109 @@ const workspaceNameRules: Array<(x: string) => string|boolean> = [
...
-    },
-  },
+      localWorkspace.value = props.workspace;
+      await store.dispatch.fetchWorkspace(props.workspace);
+      loading.value = false;
+    }
+
+    watch(() => props.workspace, () => update());
+    // eslint-disable-next-line no-return-assign
+    watch(localWorkspace, () => requestError.value = null);
```
AlmightyYakob, line 251: Rather than ignore the rule, I'd just follow it here

```suggestion
    watch(localWorkspace, () => { requestError.value = null; });
```

AlmightyYakob (changes requested):
Looks good overall, just a few suggestions

src/views/WorkspaceDetail.vue
JackWilb: See c141af6

src/views/WorkspaceDetail.vue
JackWilb: See c141af6

JackWilb:
Thanks for the suggestions, @AlmightyYakob. That clears up the code quite a bit

AlmightyYakob (approved):

PR #218, opened 2022-02-11
Only check for pending and started uploads

### LINKED ISSUE #216: Upload silently fails for example dataset
This is a movies dataset in D3 JSON format. No response after a few minutes. 

https://drive.google.com/open?id=1tGDHfAFnYhcGltE5hP2y0R-rHbePDqiq&authuser=alexander.lex%40gmail.com&usp=drive_fs

AlmightyYakob:
Closes #216 

Previously, the upload in progress spinner only checked if the status was `FINISHED`. `FAILED` is possible as well, and would result in an infinite spinner. This updates that check.


JackWilb:
On very large uploads I've also ran into times when celery would run out of memory and crash with the upload stuck in the 'started' state. This is likely not the best place to fix that, but I wanted to make a note of it

AlmightyYakob:
> On very large uploads I've also ran into times when celery would run out of memory and crash with the upload stuck in the 'started' state. This is likely not the best place to fix that, but I wanted to make a note of it

Hmm, have you tested this after the chunking change? https://github.com/multinet-app/multinet-api/pull/90

In either case, that kind of surprises me, since it'd have to be quite a large file to run out of memory. Do you remember how large the file was?

JackWilb:
> Hmm, have you tested this after the chunking change? [multinet-app/multinet-api#90](https://github.com/multinet-app/multinet-api/pull/90)

I have. The problem is that even though the inserts are chunked into Arango, they're not chunked into the celery worker, it has to open the whole file first. That's what leads to the the crash.

> In either case, that kind of surprises me, since it'd have to be quite a large file to run out of memory. Do you remember how large the file was?

On the order of hundreds of megabytes, so quite large.

JackWilb (approved):

PR #225, opened 2022-03-22
Add quick links to visualizations for networks and tables in workspace detail view

JackWilb:
No, but we have discussed this before

This adds `v-select` elements onto the network and table rows in the workspace detail view that allows a user to quickly select a visualization. The user is then redirected to the visualization.

In this PR, I had to change the structure of the appregistry to allow for both table and network visualizations, since upset only works with tables and we need to have separate lists for them.

Another additional change is that the hover no longer shows a delete and download icon per row. That's now accessed by selecting the row with the checkbox and using the icons that existed for downloading/deleting multiple tables.

Before (without hover):
image

After:
image

No


alexsb:
Love the functionality. 

I don't think the way to select the visualization is the final solution. 

I could envision a situation where you select a dataset and then hit a large "Visualize" button, then you get a menu with visualization options (and recommendations)? Maybe worth mocking this up in e.g., figma?

kirangadhave (dismissed):

waxlamp (dismissed):
Looking good in general, though dropdown-plus-button seems like the wrong UI design here. A button that says "visualize" with a dropdown whose entries go directly to the vis app would be better, or perhaps a few icon buttons inline, each of which goes right to one of the apps. (The button dropdown with direct links would likely be the best option.)

Approving because this works, but I think you should strongly consider the button dropdown design.

JackWilb:
Here's a mock up for a modal design:
https://www.figma.com/file/PVQUDeOuDocLgifgSYc0az/Client---Visualize-Menu-Mockup?node-id=0%3A1

PR #227, opened 2022-04-21
Add new mode for creating networks from existing tables

AlmightyYakob:
This adds a new mode of network creation that allows for interactive linking of existing workspace tables.


waxlamp (dismissed):

src/store/index.ts
```diff
@@ -42,7 +42,7 @@ const {
     uploads: [],
   } as State,
   getters: {
-    tables(state: State, getters): string[] {
+    tables(state: State, getters) {
```
waxlamp, line 45: Why remove the return type?

waxlamp (approved):

src/store/index.ts
AlmightyYakob: Due to changes brought in from main, the tables being returned are no longer strings, but `Table` objects. I removed the `string[]` so that the return type could be correctly inferred as `Table[]` (as is done with the similar getters below).

PR #243, opened 2022-07-19
Get table names from tables and networks for AQLWizard

### LINKED ISSUE #119: Change the "clear" button to a "clear results" button in the AQL UI
Currently the clear button clears the query, which was immediately apparent to me. ~We should either change the wording to reflect that it just clears the query text or we could update the functionality to clear out the query results, too.~

TODO:
- [x] change the button text to `clear results`
- [x] change its behavior to do so

JackWilb:
Closes #119
Fixes bug I noticed when using interface (JS objects in the AQL right sidebar)

This PR fixes a bug from the AQL sidebar and modifies the action of the `clear` button.

I noticed that there were JS objects in the AQL wizard's right sidebar, so this just applied a map the object that it relies on to get just the names from the table objects.

The clear button, now renamed `clear results` now clears the results instead of the query text, as discussed in #119.

Before:
image

After:
image

No

### DIFF
--- a/src/views/AQLWizard.vue
+++ b/src/views/AQLWizard.vue
@@ -154,9 +154,9 @@ export default {
     };
   },
   computed: {
-    nodeTables: () => store.getters.nodeTables,
-    edgeTables: () => store.getters.edgeTables,
-    networks: () => store.getters.networks,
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
This is addressed in #250

PR #246, opened 2022-08-05
Bump vue to 2.7 and ts to 4.7

JackWilb:
### Does this PR close any open issues?
No

### Give a longer description of what this PR addresses and why it's needed
This bumps vue and typescript to their newest versions. Vue 2.7 is a stepping stone to vue 3 support and with my next PRs, will make the transition nice and smooth.

In this PR, I also fixed the sass spam that would happen on build and serve.

### Provide pictures/videos of the behavior before and after these changes (optional)
N/A

### Are there any additional TODOs before this PR is ready to go?
No

PR #247, opened 2022-08-05
Migrate all components to the composition API

JackWilb:
Depends on #246 

This large PR migrates all components over to the composition API.

There were a couple of issues that I ran into when trying to do this conversion that I had to work around. The main one was that `vue-router` doesn't support the composition API on vue 2.x. As such, I have to interact with the router in cumbersome ways (through getCurrentInstance with lots of null checks). This will be fixed in a final vue 2.x compatible release of `vue-router`, when vue-router#3760 is addressed. I will make an issue to track this change.

N/A

TODOs:
- [x] Make issue for vue-router upgrade (#251)

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,5 +1,4 @@
 import 'material-design-icons-iconfont/dist/material-design-icons.css';
-import AsyncComputed from 'vue-async-computed';
 import Vue from 'vue';
 import App from './App.vue';
 import api from './api';
@@ -10,7 +9,6 @@ import store from './store';
 import './vuegtag';
 
 Vue.config.productionTip = false;
-Vue.use(AsyncComputed);
 
 oauthClient.maybeRestoreLogin().then(() => {
   Object.assign(api.axios.defaults.headers.common, oauthClient.authHeaders);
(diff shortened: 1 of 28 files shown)


AlmightyYakob:
> There were a couple of issues that I ran into when trying to do this conversion that I had to work around. The main one was that `vue-router` doesn't support the composition API on vue 2.x. As such, I have to interact with the router in cumbersome ways (through getCurrentInstance with lots of null checks).

@JackWilb I've alleviated this a bit with [862db03](https://github.com/multinet-app/multinet-client/pull/247/commits/862db0337404a82fae096053ac4734d28004abde). Does it look good to you?

src/main.ts
```diff
@@ -12,7 +11,6 @@ import './vuegtag';
 
 Vue.config.productionTip = false;
 Vue.use(AsyncComputed);
```
AlmightyYakob, line 13: This can be removed along with the entry in `package.json` as it's moot with the switch to  composition API.

JackWilb:
Looks great, let's get this merged in

AlmightyYakob (approved):

PR #249, opened 2022-08-05
Show upload errors

### LINKED ISSUE #160: Any network creation error results in "already exists" error message
The offending code is [here](https://github.com/multinet-app/multinet-client/blob/master/src/components/GraphCreateForm.vue#L89).

### LINKED ISSUE #224: Show message if upload fails
As of https://github.com/multinet-app/multinet-client/pull/218, after an upload fails, nothing will be displayed, the loader will just go await. An error message should be displayed to inform the user of what happened.

JackWilb:
Depends #247  
Closes #160 
Closes #224

This adds an error message that shows when an upload has failed. The error will show for 15 minutes and then will hide. It is not dismiss-able by the user, but that could be added as an additional feature.

image

No


src/views/WorkspaceDetail.vue
```diff
@@ -207,7 +217,8 @@ export default defineComponent({
     const tables = computed(() => store.getters.tables);
     const networks = computed(() => store.getters.networks);
     const uploads = computed(() => store.state.uploads.filter(
-      (upload) => upload.status === 'PENDING' || upload.status === 'STARTED',
+      // Find uploads with PENDING, STARTED, or FAILED status and that are less than 15 minutes old
+      (upload) => (upload.status === 'PENDING' || upload.status === 'STARTED' || upload.status === 'FAILED') && (new Date().getTime() - new Date(upload.created).getTime() < 15 * 60 * 1000),
```
waxlamp, line 221: What's the rationale behind the 15 minute cutoff?

JackWilb: We don't save any state on the browser at the moment. This seemed like a reasonable duration to show the message for. Is there a better way to handle this, maybe using local storage?

PR #250, opened 2022-08-22
Add bind vars to AQL query UI

### LINKED ISSUE #119: Change the "clear" button to a "clear results" button in the AQL UI

AlmightyYakob:
Client change for https://github.com/multinet-app/multinet-api/pull/139

Also closes #119

### DIFF
--- a/src/views/AQLWizard.vue
+++ b/src/views/AQLWizard.vue
@@ -102,9 +102,9 @@
           <v-btn
             color="error"
             text
-            @click="query=''"
+            @click="lastQueryResults = null"
           >
-            Clear
+            Clear Results
           </v-btn>
           <v-spacer />
           <v-btn
--- a/src/views/AQLWizard.vue
+++ b/src/views/AQLWizard.vue
@@ -71,36 +71,6 @@
           >
             Run Query
           </v-btn>
-
-          <v-menu
-            v-model="createTableMenu"
-            offset-y
-            :close-on-content-click="false"
-          >
-            <template #activator="{ on }">
-              <v-btn
-                class="ml-2"
-                color="secondary"
-                v-on="on"
-              >
-                Create Table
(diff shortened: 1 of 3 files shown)


src/views/AQLWizard.vue
```diff
@@ -71,36 +113,6 @@
...
-                label="Table Name"
-                append-icon="publish"
-                :error-messages="createTableErrorMessage"
-                :hide-details="!createTableErrorMessage"
-                @input="createTableErrorMessage = null"
-                @click:append="createTable"
-                @keydown.enter="createTable"
-              />
-            </v-card>
-          </v-menu>
```
JackWilb, line 103: Why remove this? It's a cool feature of this view

JackWilb (approved):
Looks good. I had one comment about some removed functionality.

Also, we will need to document @@ vs @ (or at least provide a link to arango docs). We should make an issue for that

src/views/AQLWizard.vue
AlmightyYakob: It doesn't actually work (there is no API endpoint that allows creation of a table from an AQL query), and adding it is out of scope for this PR. 

I don't know when the feature was removed / lost from the API, it obviously used to work though.

JackWilb: Gotcha. Makes sense. Happy to merge as is then

JackWilb (approved):

AlmightyYakob:
@JackWilb Can I get a re-review since updating the client library.

JackWilb (approved):

PR #255, opened 2022-08-26
Use LoginMenu from multinet-components

JackWilb:
No

This is the first step on the path to using reusable components from a central `multinet-components` repo. I have preliminarily moved the `LoginMenu` component there, and I'm using that implementation here.

To make this work, I had to change the implementation of the mdi-icons. We now use the same strategy as the other client apps, where we just get a css link in the index.html file.

Before:
image

After:
image
image

TODOs:
- [x] Fix MDI icons for avatar
- [ ] Delete old MDI icon update branch, authored by @AlmightyYakob


src/components/Sidebar.vue
```diff
@@ -35,7 +35,11 @@
         </router-link>
       </v-toolbar-title>
       <v-spacer />
-      <login-menu />
+      <login-menu
+        :store="store"
```
AlmightyYakob, line 39: Passing the entire store seems excessive and is store implementation specific. IMO you should just pass what's needed here, the callbacks for logout and fetching user info.

AlmightyYakob (approved):
One comment but otherwise lgtm. I'd be okay merging this and fixing that as a follow up

PR #259, opened 2022-09-22
Allow source/target column sharing

### LINKED ISSUE #258: Support duplicate columns in csv network builder
There's a reasonable use case where the user has one node table and one edge table. Each edge table's from and to columns could both point to the same column in that one node table. As it currently stands, this at least isn't supported in the GUI. It might not be an issue in the backend, but the GUI currently prevents this action.

This should be updated to allow for this.

AlmightyYakob:
Closes #258 

With this change, you can creating a network in the UI with the following data (not possible before)

Edge table:
```
source,target
1,2
1,3
2,3
```

Node table

```
id,name
1,jack
2,john
3,james
```

### DIFF
--- a/src/components/NetworkMultiCSVUploadForm.vue
+++ b/src/components/NetworkMultiCSVUploadForm.vue
@@ -645,8 +645,7 @@ export default defineComponent({
 
     function getOtherTableColumns(tableName: string) {
       const otherTables = visibleTableSamples.value.filter(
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
@JackWilb Can you confirm that this works as you'd like? Please test on trivially small tables since there are performance issues which aren't the goal of this PR

JackWilb (approved):
This works on a small test network. I didn't test anything large, given your comment. Thanks, Jake!

